Issue #106, opened 2026-05-26
test: pin the claude -> gateway Authorization: Bearer contract end-to-end


chengchao:
## Problem
The entire LLM path depends on the `claude` binary emitting `ANTHROPIC_AUTH_TOKEN` as an `Authorization: Bearer` header, because `apps/llm-gateway` accepts only Bearer auth. This is correct per the Claude Code LLM-gateway docs, but **no in-repo test verifies it**:
- `apps/llm-gateway/src/index.test.ts` hand-sets `Authorization: Bearer <token>`, so it proves the gateway side of the contract while asserting nothing about the client side. If a future binary version sent the credential differently (e.g. `x-api-key`), every unit test would still pass and every live turn would `401`.
- The gateway's request-body streaming path (`duplex: "half"`) is only exercised with a mocked `fetch` (finding #10), so real streaming is unverified.

## Action
- [ ] Add a real end-to-end check that drives the actual `claude` binary through the gateway and asserts a real completion (naturally folds into the E2B integration rework — see the "rework E2B integration + probe scripts" issue).
- [ ] Optionally, a focused contract test that captures the binary's outgoing headers and asserts `Authorization: Bearer` is present.

## Refs
Code-review findings #4 (binary->gateway contract / false confidence) and #10 (duplex streaming untested) from the LLM-gateway re-architecture review.

PR #107, opened 2026-05-26
feat(sandbox): route agent LLM calls through an llm-gateway control plane

chengchao:
## Summary
Moves provider credentials out of the E2B sandbox. The sandboxed agent is treated as untrusted; it no longer holds `ANTHROPIC_API_KEY`. Instead:

- **chat-api** mints a short-lived, HMAC-signed bearer token per turn (bound to `{userId, sandboxId, requestId}`) and sends `llm_base_url` + `llm_token` in the `/turn` body.
- **sandbox-daemon** sets these as `ANTHROPIC_BASE_URL` + `ANTHROPIC_AUTH_TOKEN` on the agent process — the daemon holds no provider key.
- **llm-gateway** (new, stateless service) verifies the token, injects the real `x-api-key`, and proxies to Anthropic.

Net effect: the real key lives in exactly one small service, and a compromised/prompt-injected agent has nothing worth stealing beyond a 10-minute, single-user token.

```
chat-api (mint) ──llm_token──► daemon ──ANTHROPIC_AUTH_TOKEN──► claude
                                                                  │ Authorization: Bearer
                                                                  ▼
                                            llm-gateway (verify → inject x-api-key) ──► api.anthropic.com
```

## What's included
- `packages/llm-token` — shared mint/verify (HMAC; payload-shape validated so a signed-but-malformed payload can't bypass expiry).
- `apps/llm-gateway` — token-gated transparent proxy; Anthropic-shaped error on upstream transport failure; trimmed/case-insensitive Bearer parse; validated `GATEWAY_PORT`; base-URL trailing-slash normalization.
- chat-api: token minting + turn-body fields; **dropped `ANTHROPIC_API_KEY`** and the unused OpenAI/DeepSeek/ai-sdk deps.
- sandbox-daemon: forwards the token into the agent env.
- Tests pinning the `mint → forward → agent-env → verify` seams, including an assertion that **no provider key reaches the agent**.
- `AGENTS.md`, `Dockerfile`, `compose.yaml` updated.

## Tests
`llm-token 8/8 · llm-gateway 3/3 · sandbox-daemon 87/87 · chat-api 37/37`; gateway typechecks clean; biome clean on changed source.

## Deploy prerequisite
`LLM_GATEWAY_PUBLIC_URL` (chat-api) must resolve to an address **reachable from inside the E2B sandbox**, and `ANTHROPIC_API_KEY` + `LLM_TOKEN_SECRET` must be set on `llm-gateway` (the latter shared with chat-api). `ANTHROPIC_API_KEY` is no longer needed by chat-api or the daemon.

## Deferred follow-ups
- #103 — rework the E2B integration + probe scripts (currently broken under the new arch; in-sandbox gateway decided)
- #104 — decide token TTL vs. claim-binding enforcement
- #105 — prevent the bearer token leaking into `daemon.log`
- #106 — pin the claude → gateway `Authorization: Bearer` contract end-to-end

### DIFF
--- a/apps/llm-gateway/src/index.ts
+++ b/apps/llm-gateway/src/index.ts
@@ -43,7 +43,9 @@ const RESPONSE_DROP_HEADERS = new Set([
 
 export const app = new Hono();
 
-app.get("/health", (c) => c.json({ status: "ok" }));
+// GET and HEAD so load-balancer / k8s probes (which often use HEAD) don't fall
+// through to the token-gated proxy and 401.
+app.on(["GET", "HEAD"], "/health", (c) => c.json({ status: "ok" }));
 
 // Token-gated proxy for the messages endpoints. Catch-all so a trailing-slash
 // base URL (`…//v1/messages`) still routes here and gets normalized below.
--- a/apps/llm-gateway/src/index.ts
+++ b/apps/llm-gateway/src/index.ts
@@ -72,10 +72,12 @@ async function proxyToAnthropic(c: Context) {
 		);
 	}
 
-	// Collapse duplicate slashes (a trailing-slash base URL yields `//v1/messages`)
-	// before the scope check and forwarding.
+	// Normalize the path before the scope check / forwarding: collapse duplicate
+	// slashes (a trailing-slash base URL yields `//v1/messages`) and drop a single
+	// trailing slash (`/v1/messages/` → `/v1/messages`) so exact-match still holds.
 	const url = new URL(c.req.url);
-	const path = url.pathname.replace(/\/{2,}/g, "/");
+	let path = url.pathname.replace(/\/{2,}/g, "/");
+	if (path.length > 1 && path.endsWith("/")) path = path.slice(0, -1);
 	if (!ALLOWED_PATHS.has(path)) {
 		return c.json(
 			{
--- a/apps/llm-gateway/src/index.ts
+++ b/apps/llm-gateway/src/index.ts
@@ -1,0 +1,124 @@
+import { verifyLlmToken } from "@mymemo/llm-token";
+import { type Context, Hono } from "hono";
+import { gwEnv } from "./env";
+
+/**
+ * LLM gateway — the control plane for sandboxed agents.
+ *
+ * Sandboxed agents point the Claude binary at this service via ANTHROPIC_BASE_URL
+ * and authenticate with a short-lived ANTHROPIC_AUTH_TOKEN (Bearer). The agent
+ * therefore holds no provider key: we validate the token, inject the real
+ * `x-api-key`, and stream the upstream response straight back.
+ *
+ * The gateway is a transparent, token-gated reverse proxy: it forwards whatever
+ * path/method/headers the client sends (so SDK features and any Anthropic
+ * endpoint keep working) and only rewrites auth. Everything except /health is
+ * proxied.
+ */
+
+// Client request headers we must NOT forward upstream: authorization is replaced
+// by x-api-key; host/content-length/connection are recomputed by fetch (the body
+// is re-streamed with chunked encoding).
+const REQUEST_DROP_HEADERS = new Set([
+	"authorization",
+	"host",
+	"content-length",
+	"connection",
+]);
+
+// Response headers fetch already decoded for us; forwarding them would mislead
+// the client into decompressing again or mismatching the streamed length.
+const RESPONSE_DROP_HEADERS = new Set([
+	"content-encoding",
+	"content-length",
+	"transfer-encoding",
+	"connection",
+]);
+
+export const app = new Hono();
+
+app.get("/health", (c) => c.json({ status: "ok" }));
+
+// Token-gated transparent proxy for everything else.
+app.all("*", proxyToAnthropic);
+
+function bearerToken(c: Context): string {
+	const auth = c.req.header("authorization")?.trim() ?? "";
+	const match = /^Bearer\s+(.+)$/i.exec(auth);
+	return match?.[1] ?? "";
+}
+
+async function proxyToAnthropic(c: Context) {
+	const claims = verifyLlmToken(bearerToken(c), gwEnv.LLM_TOKEN_SECRET);
+	if (!claims) {
+		return c.json(
+			{
+				type: "error",
+				error: {
+					type: "authentication_error",
+					message: "invalid or expired session token",
+				},
+			},
+			401,
+		);
+	}
+
+	// Cost-cap / metering hook: claims.userId identifies the end user, and
+	// X-Claude-Code-Session-Id aggregates a session without parsing the body.
+
+	const headers = new Headers();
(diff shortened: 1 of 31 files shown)


chengchao:
## ✅ Live end-to-end smoke passed

Ran `bun run smoke:gateway` (added in 7efdfc3) against the **real Anthropic API** — boots the gateway locally, mints a token, and exercises the credential path:

```
Layer A — gateway ⇄ Anthropic
  ✓ valid token streams a completion — "pong"
  ✓ garbage token → 401
  ✓ expired token → 401
  ✓ missing token → 401
Layer B — claude binary → gateway (Authorization: Bearer)
  ✓ claude completes a prompt through the gateway — "pong"
```

**Layer B closes the #106 risk:** the real `claude` binary sends `ANTHROPIC_AUTH_TOKEN` as `Authorization: Bearer`, the gateway verifies it, injects the real `x-api-key`, and a completion streams back. That was the one assumption no unit test could prove (the gateway unit test hand-sets the header). It's now verified live, along with the 401 auth-gating and the streaming/duplex request path.

### Remaining before merge (process/ops, not unknowns)
- Review/approval (no CI in this repo, so human review is the only gate).
- Deploy prereq: `llm-gateway` deployed with `ANTHROPIC_API_KEY` + `LLM_TOKEN_SECRET`, and `LLM_GATEWAY_PUBLIC_URL` set to an address **reachable from inside the E2B sandbox** (the smoke ran the gateway on localhost).
- Optional: decide #104 (token TTL vs. binding) if turns can run long.

Follow-ups tracked: #103 (E2B script rework), #104, #105, #106.

chengchao (commented):
Inline notes from review. Architecture and the security-sensitive code (token mint/verify, header rewriting) are correct and well-tested — comments below are about bounding the now-internet-reachable gateway and a few framing/config nits, not blockers to the approach.

apps/llm-gateway/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+				error: {
+					type: "authentication_error",
+					message: "invalid or expired session token",
+				},
+			},
+			401,
+		);
+	}
+
+	// Cost-cap / metering hook: claims.userId identifies the end user, and
```
chengchao, line 94: **Cost/abuse surface — closer to blocking than deferred.** This gateway must be reachable from inside E2B sandboxes (effectively the public internet) and its only auth is the HMAC token. Within the 10-min window a prompt-injected agent can pick any model + `max_tokens` and burn unbounded budget on the company key — and because the route is `app.all("*")` it can hit *any* Anthropic endpoint (batches, files, …), not just `/v1/messages`. The `userId`/`sandboxId` binding is minted but never enforced here (#104), and there's no rate/spend cap — only this comment. A coarse per-token request cap feels like it should land with this PR, not after.

apps/llm-gateway/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+	"transfer-encoding",
+	"connection",
+]);
+
+export const app = new Hono();
+
+app.get("/health", (c) => c.json({ status: "ok" }));
+
+// Token-gated transparent proxy for everything else.
+app.all("*", proxyToAnthropic);
```
chengchao, line 52: Minor: `app.all("*")` also matches non-GET methods on `/health`, so e.g. `POST /health` is proxied upstream with the real key. Harmless, but slightly surprising.

apps/sandbox-daemon/child-spawn.ts
```diff
@@ -268,11 +274,14 @@ export interface SpawnAgentResult {
 export async function spawnAgent(
 	input: SpawnAgentInput,
 ): Promise<SpawnAgentResult> {
-	const { onEvent, ...config } = input;
+	const { onEvent, llmBaseUrl, llmToken, ...config } = input;
 	ensureClaudeProjectsDir();
 	const proc = Bun.spawn(buildAgentSpawnArgv(config.cwd), {
 		env: {
-			ANTHROPIC_API_KEY: process.env.ANTHROPIC_API_KEY ?? "",
+			// The agent holds no provider key — it talks to the LLM gateway, which
```
chengchao, line 281: The win is real but scoped to *this* (agent.js) process. `DATABASE_URL` still lives in the sandbox VM (daemon + sync.js env), and full DB creds are a bigger prize than an Anthropic key — so the PR summary's "nothing worth stealing beyond a 10-min token" overstates it. Worth a line in the threat model noting that bwrap process isolation is the load-bearing control keeping the agent away from `DATABASE_URL`.

apps/sandbox-daemon/child-spawn.ts
```diff
@@ -268,11 +274,14 @@ export interface SpawnAgentResult {
...
+			// injects the real key. ANTHROPIC_AUTH_TOKEN is sent as a Bearer header.
+			ANTHROPIC_BASE_URL: llmBaseUrl,
+			ANTHROPIC_AUTH_TOKEN: llmToken,
```
chengchao, line 284: `LLM_TOKEN_SECRET` is now a shared high-value secret across two services, and this bearer token is the entire blast-radius control for the public gateway. #105 (token leaking into `daemon.log`) is directly exploitable against it — a leaked token = company-key access until expiry — so it feels like it should land *with* this PR rather than after.

compose.yaml
```diff
@@ -8,6 +8,25 @@ services:
       - 3000:3000
     env_file:
       - ./apps/chat-api/.env
+    depends_on:
```
chengchao, line 11: `depends_on` without `condition: service_healthy` only orders container *start*, not readiness — chat-api can boot before the gateway accepts connections. Low severity (per-turn calls, not startup), but if you want the ordering to mean anything, add a `/health` healthcheck and `condition: service_healthy`.

compose.yaml
```diff
@@ -8,6 +8,25 @@ services:
...
+  # agents reach it via ANTHROPIC_BASE_URL with a short-lived bearer token.
+  # Note: LLM_GATEWAY_PUBLIC_URL (set for chat-api) must resolve to an address
+  # reachable from inside the E2B sandbox, not just this compose network.
+  llm-gateway:
+    container_name: llm-gateway
+    build:
+      context: .
+      dockerfile: apps/llm-gateway/Dockerfile
+    ports:
+      - 8081:8081
```
chengchao, line 27: `8081:8081` is hardcoded to match the `GATEWAY_PORT` default. Fine as-is, but if `apps/llm-gateway/.env` overrides `GATEWAY_PORT`, the published port silently breaks. A comment here would save someone a confusing debug session.

bun.lock
```diff
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
```
chengchao, line 2: `"configVersion": 0` was dropped from the lockfile in this hunk — confirm that's an intended format change from your Bun version and not a stray local-Bun artifact that'll churn for other contributors.

PR #141, opened 2026-06-19
fix(workspace-store): atomic docs-manifest writes under concurrency (MYM-37)

chengchao:
## What

`writeDocsManifest()` wrote to a **fixed** sibling temp path `${target}.tmp`, then renamed it into place for atomicity. chat-api does **not** serialize turns per conversation (`chat.controller.ts` derives `conversationId` from the body; `runSandboxChat()` takes no lock), and the durable root `users/{userId}/conversations/{conversationId}/docs/` is shared across turns. Two concurrent turns for the same conversation therefore wrote the **same** `manifest.json.tmp`: `writeFileSync` truncates, so the writers clobbered each other's temp bytes and their renames raced — producing a lost update, a byte-mixed manifest, or an `ENOENT`-on-rename crash.

This PR makes the atomic-write path safe under concurrency.

## Why

The durable docs manifest is the record of a conversation's hydrated working set. A corrupted or partially-overwritten `manifest.json` is a fail-closed error (`readDocsManifest` throws on malformed input), so a torn write breaks the next turn for that conversation. MYM-9's acceptance criteria called for writes to avoid corruption; this closes the concurrency gap that slipped through.

## Changes

- **Unique temp name per write** — `${target}.${pid}.${random}.tmp`. Concurrent writers never share a temp path; each renames its own complete file in (last-writer-wins, never torn). Per-pid alone is unsafe across containers (shared PID namespaces), so the random token is required for a shared volume.
- **Orphan temp sweep** (`sweepStaleTemps`, 60s threshold) — unique names defeat the old self-healing reclamation, so a writer that crashes between write and rename would leak its temp forever. Best-effort, age-based so a concurrent live writer's fresh temp is never swept.
- **Scoped docstring** — the guarantee is corruption-freedom only (last-writer-wins, not a merge). Read-modify-write callers (`upsertDocsManifestEntry`, future hydrate→sync) can still drop a concurrent change and must serialize themselves; documented on both `writeDocsManifest` and `upsertDocsManifestEntry`.
- Mirrored across the **chat-api** and **sandbox-daemon** copies for parity (daemon is lower risk: turn-locked, one sandbox per conversation).

## Tests

- New concurrency test (both apps): 8 subprocesses write the same `docs/` dir; the final `manifest.json` must be valid JSON and exactly one writer's complete output. **Verified to fail on the old fixed-temp code** and pass on the fix.
- New stale-temp-sweep test: an orphaned (backdated) temp is reclaimed while a fresh one is spared. **Verified to fail if the sweep call is removed.**
- Full suites green: **daemon 63/63, chat-api 84/84**; lint clean on all touched files.

## Reviewer notes

- The logical **lost update** under concurrent read-modify-write is documented but intentionally not eliminated — no concurrent RMW caller exists yet in chat-api, and the daemon relies on its per-turn lock. If such a path is added later, serialize manifest mutations per conversation.

Closes MYM-37. Related: MYM-9, MYM-10.

### DIFF
--- a/apps/chat-api/src/features/workspace-store/docs-manifest.test.ts
+++ b/apps/chat-api/src/features/workspace-store/docs-manifest.test.ts
@@ -1,0 +1,80 @@
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { mkdtempSync, readdirSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { readDocsManifest } from "./docs-manifest";
+
+let docsDir: string;
+
+beforeEach(() => {
+	docsDir = mkdtempSync(join(tmpdir(), "chat-api-docs-manifest-test-"));
+});
+
+afterEach(() => {
+	rmSync(docsDir, { recursive: true, force: true });
+});
+
+describe("writeDocsManifest concurrency", () => {
+	// chat-api does not serialize turns per conversation, so two concurrent turns
+	// can call writeDocsManifest against the same durable `docs/` dir at once.
+	// This proves the final manifest.json is always one writer's complete output,
+	// never a torn or byte-mixed file (which a shared fixed temp name would cause).
+	it("concurrent writers never produce a torn or byte-mixed manifest", async () => {
+		const WRITERS = 8;
+		// A shared fixed temp name makes concurrent writers collide on one path:
+		// they truncate/clobber each other's in-flight bytes and their renames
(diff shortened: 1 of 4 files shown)


chengchao:
Closing as stale. Opened 2026-06-19; both modules it patches have since been removed: `apps/sandbox-daemon/` was deleted with the prototype path (ADR-0002, split runtime), and `apps/chat-api/src/features/workspace-store/` no longer exists. All four changed files are absent from main, and the PR is CONFLICTING. Nothing here is salvageable against the current tree.

PR #372, opened 2026-07-24
feat(live-text): add producer-buffered relay

chengchao:
## Summary

- define the producer-buffered Live Stream relay contract from ADR-0014 with equivalent in-memory and Redis pub/sub implementations
- fail-stop producers on the first live publish or per-Run cap failure, reject later appends, and end attached or attaching readers with a coded relay-failure outcome
- return a distinct prompt `aborted` attach outcome and deterministically verify the backlog/live ordinal-discard seam
- consolidate shared identifier validation, expose a named subscription type, and replace prose-matched relay lifecycle failures with coded errors

## Why

Issue #367 establishes and proves the replacement relay before #369 cuts production traffic over to it. Latching failures keeps backlog and live views from diverging and gives callers an explicit path to Recovering.

## Impact

There is no production wiring change in this PR. The existing retained Redis Stream remains active until the follow-up cutover. Relay factory URL validation remains assigned to #369. The `LiveStreamStoreError` rename and legacy-code removal are deliberately tracked on #371.

## Validation

- `bun test`: 797 passed, 20 environment-gated skipped, 0 failed
- strict TypeScript checks passed for all five workspaces
- `@mymemo/live-text`: 42 tests passed against in-memory, real Redis pub/sub, and the retained-Stream regression contract
- formatting and final two-axis standards/spec review completed with no remaining findings

Closes #367


chengchao (commented):
Two-axis review (/code-review) against #367 and ADR-0014. All 18 contract tests pass at f671117 against in-memory and a spawned real Redis, and the wire-level ADR constraints check out (ordinals never leave the transport, per-reader reply channels, payload-free telemetry, caps preserved, glossary vocabulary correct). Four findings are blocking; the rest are triaged below.

## Blocking — the relay's own contract (#367 acceptance criteria)

1. **Failed live publish desynchronizes backlog and live views.** In `live-stream-relay.ts` `#publish`, `#events.push` precedes `transport.publish` and is not rolled back on failure, so an event can exist in every future backlog reply yet never appear on the live channel — already-attached readers stall on that ordinal forever with no error. Fix with ADR-0014's fail-stop posture, not fault-tolerance: on a failed publish, latch the producer failed, reject further appends, and end every attached reader with an explicit relay-failure outcome so callers route to Recovering. Never keep growing a backlog live readers can no longer converge with. (For contrast: vercel/resumable-stream halts its whole loop on an unhandled publish rejection — everything hangs, but backlog and live at least freeze *together*. This PR keeps the optimism without that accidental consistency.)

2. **Cap crossing starves attached readers of the terminal.** `publishTerminal` is subject to the same byte/event caps, so a Run at cap can never deliver its terminal — readers hang with no in-band closure, against "the terminal event is delivered to attached consumers and ends their sequences." Falls out of fix 1 if cap-crossing takes the identical latched-failure path (readers get an explicit failure end; the Outcome comes from history). Optionally also exempt the terminal event from the caps.

3. **Abort conflation.** `attach` on an already-aborted signal resolves `{ outcome: "no_producer" }` while telemetry records `aborted`, and an abort during the wait doesn't settle the attach at all (it runs the full timeout). The spec requires a *distinct* no-producer outcome; add a distinct `aborted` outcome that settles promptly.

4. **Seam arithmetic not deterministically exercised.** The "joins a concurrent backlog and live tail" contract test races `attach` against appends via `Promise.all`, but nothing forces the load-bearing branch — a live event buffered pre-reply with ordinal < `reply.count` that must be discarded (`settleAttached`). The suite can pass every run without that path executing, especially in-memory where the reply settles synchronously. Gapless-by-arithmetic is the entire advantage over upstream's gapless-by-synchronicity; force this branch deterministically.

## Cheap to fold in while the package is newborn (non-blocking)

- Deduplicate `positiveInteger` (relay + store) and the triplicated run-id validation (`validateRunId` inlines `128`/pattern beside the store's `MAX_RUN_ID_LENGTH`/`RUN_ID_PATTERN` and `live-stream-events.ts`'s own copy).
- Name the `{ close(): Promise<void> }` subscription handle (repeated inline five times across three files).
- Replace bare `Error` prose messages that the contract tests string-match (`toThrow("requires a terminal")`, `"already published"`) with coded errors, matching the pattern one import away.
- `encodeEvent`/`decodeEvent` in `live-stream-relay.ts` re-create `TextEncoder`/`TextDecoder` per call beside the module-level codecs in `live-stream-events.ts`.

## Deferred deliberately (do not fix here)

- **Relay factory URL validation** (`createRedisLiveStreamRelay` accepts any `url`) → #369, where production wiring happens; secure-URL enforcement lives at env resolution today and the contract tests legitimately need loopback `redis://`.
- **`LiveStreamStoreError` naming** — the relay throwing a *store* error whose codes include `invalid_cursor` brushes the ADR's "there is no store" claim and the CONTEXT.md avoid-list, but six of its ten codes die with the store → note added to #371 territory; renaming the shared type while the store still uses it is churn in the wrong direction.
- **Scope-creep items** — the `redis-live-stream-store.ts` → `live-stream-events.ts` extraction is behavior-preserving motion #371 deletes anyway; `testLimits` is clamped to real caps and the 8 MiB test can't do without it. Both fine to keep.

chengchao (commented):
Second-round two-axis review of a79622c (both commits). **Verdict: mergeable on both axes — all four round-one blockers verified fixed.** Suite run at the SHA in a scratch worktree: 42/42 pass; the contract file (13 cases × 2 instantiations) passed 3 consecutive reruns against real spawned Redis.

## Round-one blockers — all resolved

1. **Fail-stop on publish failure** ✔ — `#fail()` latches, later appends reject `producer_failed`, a `relay_failed` envelope ends attached readers, and backlog requests answer `relay_failed` so mid-handshake and later attaches resolve explicitly instead of hanging.
2. **Cap crossing** ✔ — bytes, events, and the terminal all take the same latch path; "fails readers when a terminal event crosses the event cap" proves readers get `relay_failed` instead of hanging.
3. **Abort semantics** ✔ — pre-aborted signals return a distinct `aborted` outcome; mid-wait abort settles promptly; listener/settler wiring is synchronous.
4. **Seam determinism** ✔ — `beforeBacklogReply`/`afterLiveEventBuffered` hooks force the ordinal-below-reply-count discard branch deterministically, exactly-once asserted, on real Redis too.

## Two notes worth a line each

- **Acceptance-criterion supersession:** #367's verbatim criterion said buffer bounds surface "without tearing down already-attached readers." The round-one fix mandate deliberately supersedes that — readers now end with explicit `relay_failed`, which is the correct ADR-0014 behavior. Suggest one line in the PR description noting the criterion is superseded, so the checkbox isn't read as violated.
- **Unspecified edge:** producer `close()` with a still-attached reader leaves that reader's iterator hanging until the consumer abandons it. Out of scope for #367; #369's route layer should either always abort readers it owns or this gets an explicit outcome later.

## Non-blocking polish (fine to ride along now or land with #369)

- `decodeEvent`/`encodeEvent` in `live-stream-relay.ts` still construct `TextDecoder`/`TextEncoder` per call on the per-event hot path (the module-level codec pattern from `live-stream-events.ts` applies).
- The settle shape in `attach()` repeats five times (`settleAborted`, `settleAttached`, `settleRelayFailed`, timeout callback, IIFE catch) — one `settle(result)` helper would collapse them.
- `parseLiveEnvelope`/`parseBacklogReply` duplicate a dead-branch ternary for non-negative-safe-integer parsing — extractable.
- `validateLiveStreamRunId`'s `qualifier` special-case exists only to reproduce legacy message strings nothing string-matches; one message would do, and store-side `validateRetryId` is now a one-line pass-through.
- The relay still throws `LiveStreamStoreError`, and the code/reason unions now mix legacy-store and relay vocabulary (`invalid_cursor` beside `producer_failed`) — acceptable while the store coexists; prune at #344/#371 as already noted there.

Standards: ADR-0014 constraints all verified (ordinals never on the wire, payload-free telemetry, per-reader reply channel, shared live channel, fail-stop latch, caps); CONTEXT.md vocabulary correct; no hard violations. Spec: no missing criteria, no scope creep, no new defects from the fix commit.

PR #373, opened 2026-07-24
feat(chat-api): drop AG-UI wire cursors (#368)

chengchao:
## Summary

- emit data-only SSE frames from the original Run response and reconnect endpoint
- ignore residual `Last-Event-ID` headers and rebuild every active Run attachment from the retained stream's beginning
- remove `activeRun.lastEventId` from the history type, Postgres projection, and HTTP wire
- align the smoke client, retained-store integration expectations, and README with the cursor-free contract

## Why

Issue #368 is the wire-contract prefactor for ADR-0014. The web client now rebuilds active Runs on reconnect, so chat-api no longer needs to expose Redis Stream cursors or reject stale browser headers before issue #369 cuts production traffic over to the producer-buffered relay.

## Impact

Clients receive no SSE `id` fields and every reconnect rebuilds the active Run from `RUN_STARTED`. A residual `Last-Event-ID` header is harmless and ignored. The existing 503-before-first-event and 410-with-history-recovery contracts are unchanged. The retained Redis Stream remains the internal transport until #369.

## Validation

- `bun test`: 794 passed, 20 environment-gated skipped, 0 failed
- chat-api strict TypeScript check passed
- Biome formatting/lint and `git diff --check` passed
- parallel standards/spec review completed; the one standards vocabulary finding was fixed and the spec review had no findings

Closes #368


chengchao (commented):
Two-axis review (/code-review) of e7e2d3b against #368 and ADR-0014. **Verdict: one hard finding to fix — a two-line doc correction — then mergeable. Spec axis is fully conformant.**

## Standards

**Hard violation (fix before merge):**

1. **AGENTS.md still documents cursors on the AG-UI wire.** The commit updated README.md but not AGENTS.md: item 3 still reads "emitting standard AG-UI JSON in SSE `data` with the Redis cursor in `id`", and item 8 still says "standard cursor-bearing AG-UI `RUN_STARTED`…" (lines 113/118 at e7e2d3b). ADR-0014 is now the governing description of this surface ("no cursors on the wire"), CONTEXT.md's Live-Stream avoid-list includes "replay cursor", and the Surgical Changes rule applies — this change orphaned that text and should remove it.

**Judgement calls (non-blocking):**

2. The from-zero read is a magic `""` hardcoded at both read sites in `conversations.route.ts` (`streamAgUiRun`, `waitForAndStreamAgUiRun`). Acceptable interim since the retained store's `read(runId, cursor, signal)` signature survives until #369 — a named constant would help, or the param simply dies at the cutover.
3. The "reject any `id:` line" guard is hand-rolled in three independent SSE parsers (route test, e2e, smoke). Deliberate suite independence; noted, not asked.

**Orphan check otherwise clean:** cursor imports, both 400 branches, and the terminal-plus-cursor 204 branch fully removed from the route; `activeRun.lastEventId` removed consistently across type, projection, and wire; `parseLiveStreamCursor` correctly survives inside the retained store until #369; history pagination cursors rightly untouched (unrelated opaque-page concept, ADR-0012).

## Spec

All six acceptance criteria verified; no scope creep; nothing implemented-but-wrong:

- Id-less frames pinned across all streaming tests (`parseAgUiSse` throws on any `id:` line).
- Last-Event-ID ignored, never rejected — malformed case pinned in the route test, well-formed Redis-shaped case pinned in e2e (`9999999999999-0` → 200 + full replay). Acceptable coverage split.
- Replay-from-zero holds on both the POST response and the GET events endpoint.
- 503-before-first-event and 410-with-recovery contracts untouched. With the 204 branch gone, a terminal stream with no terminal event to replay falls to 410-recovery — already pinned by a pre-existing test, and no client-visible regression since the 204 only ever fired with a nonempty cursor, which no longer exists.
- Postgres-first ownership authorization before any Redis read is intact (ADR-0012 non-transport rule).
- No worker or live-text changes; the #344-deleted legacy path is not resurrected.

Process note: blockers were respected — #344 closed and mymemo-web#88 (client rebuild-on-reconnect) closed before this landed, so no cursor-dependent client can receive the id-less stream.

PR #374, opened 2026-07-24
Cut Live Stream delivery over to relay

chengchao:
## Summary

- cut chat-api and agent-worker over from retained Redis Streams to the producer-buffered Redis pub/sub relay
- rebuild active attachments from the full producer backlog plus live tail, with queued keepalives and terminal recovery through Conversation history
- preserve durable Run execution across relay failures, enforce buffer caps, bound Redis operations, and release producers on every terminal or ownership-loss path
- update the real-Redis integration fixture, smoke verifier, telemetry/runbook guidance, and infrastructure descriptions for the relay contract

## Why

Redis is now only the ephemeral cross-process relay. Active Run backlog lives in the claiming worker's memory, while Postgres remains authoritative for Run discovery, Outcomes, and permanent history. This removes retained per-Run Redis storage and its cursor/TTL lifecycle without changing the public AG-UI event vocabulary.

## User and developer impact

Active reconnects rebuild from the complete backlog without gaps or duplicates. Terminal or failed Live Streams direct clients to Conversation history. Relay outages continue to degrade only live delivery and cannot stall or change durable Run execution.

## Validation

- `bun test`: 783 passed, 20 environment-gated skipped, 0 failed
- real Redis relay contract: 27 passed
- worker/chat focused relay tests: 71 passed
- smoke/deploy verification: 38 passed
- agent-worker, chat-api, and live-text TypeScript checks
- Biome checks and Terraform formatting
- standards and specification review passes

Closes #369


chengchao (commented):
Two-axis review (/code-review) of 4209c5c against #369 and ADR-0014. **Verdict: one contained defect to fix — the reader-abort wiring in `streamAgUiRun` — then mergeable.** Everything load-bearing in the cutover verified correct.

## Spec

**Fix before merge:**

1. **Reader-abort on route give-up is half-wired.** The edge deferred from the PR #372 review to this ticket — readers the route owns are aborted when the route gives up — is wired correctly in `waitForAndStreamAgUiRun` (`attach(runId, readSignal)` + `finally` abort) but not in `streamAgUiRun`: `respondWithAgUiRun` attaches with `c.req.raw.signal`, while the give-up paths (keepalive failure, `finally`) abort `readAbort.controller` — dead wiring left from the old `liveStreamReader.read(..., readAbort.controller.signal)` call, no longer connected to anything. Since `iterator.return()` cannot preempt a pending `next()`, a keepalive-failure give-up leaves the SSE callback and relay subscriptions lingering until the producer terminates or closes. Fix: attach with (or abort) the same controller the give-up paths use, as the wait path already does.

**Design deviation, flagged but accepted:** the deferred edge was largely solved relay-side — `BufferedLiveStreamProducer.close()` now broadcasts `relay_failed` to attached readers when closed without a terminal, reversing #372's "close leaves readers hanging by design." Defensible (it bounds the lingering above and strengthens fail-stop), but note it moves a responsibility the ticket assigned to the route into the relay; with the route-side fix above, both layers cover it.

**Verified correct, explicitly:** terminal publishes strictly after the Postgres terminal transition (fence rejection → no publish); producer released on every exit path — clean/error/interrupt terminals via try/finally, ownership loss, and shutdown (drain then relay close); fail-stop latch is fenced via `markLiveStreamFailedTx` and never fails the Run; buffer-cap overflow fails live delivery only; 503-versus-410 discrimination unchanged; terminal attach serves history with no post-terminal replay; ownership authorization precedes every Redis access; `RunProcessContext.appendLiveEvent` and Run supervision untouched; app tests run on `createInMemoryLiveStreamRelay` and the worker's hand-rolled fake is deleted; no production code path consumes the retained store; no #370 failure-matrix or #371 deletion work smuggled in; Terraform edits are description-only; the ADR-0014 edit is a faithful status update.

## Standards

No hard violations. The prior AGENTS.md finding is fixed (current wire described cursor-free; "Redis stores no stream content"). Judgement calls, non-blocking:

1. **Orphaned telemetry vocabulary** — `live-stream-telemetry.ts` retains store-era operations (`acquire`/`append`/`finalize`/`read_wait`/`refresh`) and results (`consumer`/`producer`/`finalized`) whose only recorders this diff deleted; the telemetry test still exercises `acquire`. Prune here or note deferral to #371.
2. **Duplicated route sequence** — `waitForAndStreamAgUiRun`'s inner attach → first-read → done-check → write mirrors `streamAgUiRun`, the tri-condition give-up check appears twice, and entry does back-to-back `getRun` calls. Worth collapsing when touching the file for the abort fix.
3. Comment drift: `ag-ui-text-stream.ts` still says "At most one Redis append is in flight" — it's a producer-buffer append now.
4. `e2e/synthetic-worker.ts` `faultOrdinal` magic positions (2/4/8); test-only, tolerable.

Process note: frontier discipline held — #367/#368 closed and PR #373 merged before this landed.

PR #375, opened 2026-07-24
fix(chat-api): end attached readers on silent producer death (#370)

chengchao:
## Summary

- exercise the production RunLoop and chat-api attach route against one PGlite database and disposable real Redis
- deterministically cover queued attachment, attached-reader producer death, both terminal attach-race outcomes, buffer-cap overflow, and Redis unavailability
- stop attached API streams when durable Run state proves that Conversation-history recovery is required
- share the Redis lifecycle and cursor-rejecting SSE parser, and use semantic event-type/terminal failure hooks instead of ordinals
- run the PGlite/Redis matrix in the general test job; reserve the Postgres integration job for the real-Postgres suite

## Why

ADR-0014 makes worker memory the active Live Stream backlog and Postgres the durable recovery plane. Issue #370 requires deterministic cross-service proof that relay failure affects only live delivery, never durable Run execution or Conversation-history recovery.

A producer can disappear without publishing a relay-failure notification. The API therefore checks durable Run state at the existing keepalive cadence while a reader is attached and closes the incomplete live response once the Run is terminal or marked degraded, allowing the next attach to recover from history.

## Impact

- attached live responses now converge after silent producer death instead of waiting indefinitely on an otherwise healthy Redis subscription
- no database schema, authorization, admission, durable Run-processing, or public recovery contract changes
- Redis client naming and relay lifecycle hooks are opt-in and used only by deterministic integration tests

## Validation

- `bun run test`: 788 passed, 0 failed
- affected relay, chat-api route, and matrix suites: 90 passed, 0 failed
- ADR-0014 real-Redis failure matrix: 6 passed, 0 failed
- strict TypeScript checks passed for live-text, agent-worker, and chat-api
- standalone synthetic worker bundle passed
- Biome and `git diff --check` passed

Closes #370

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y ripgrep redis-server
       - run: bun install --frozen-lockfile
       - run: bun run test
+      - run: bun test e2e/relay-failure-matrix.integration.test.ts
 
   # Build the final production-pruned image and exercise the SDK's native CLI
   # inside it. This needs no runtime credentials and catches both optional-dep
@@ -39,9 +40,9 @@ jobs:
 
   # Split-runtime AG-UI integration: runs chat-api plus a deterministic
   # test-only worker against real Postgres and Redis, so atomic admission,
-  # durable completion, and the retained per-Run stream are exercised on every
-  # PR without live provider credentials. Task 9.7 owns the production SDK
-  # path's credentialed smoke.
+  # durable completion, and relay-backed streaming are exercised on every PR
+  # without live provider credentials. Task 9.7 owns the production SDK path's
+  # credentialed smoke.
   # `e2e/integration.test.ts` is gated on AGENT_DATABASE_URL.
   integration:
     runs-on: ubuntu-latest
--- a/apps/agent-worker/src/run-live-stream.test.ts
+++ b/apps/agent-worker/src/run-live-stream.test.ts
@@ -4,6 +4,7 @@ import {
 	createInMemoryLiveStreamRelay,
 	decodeAgUiLiveStreamEvent,
 	type LiveStreamMetricEvent,
+	type LiveStreamRelayOptions,
 	type LiveStreamTelemetry,
 } from "@mymemo/live-text";
 import type { WorkerLogger } from "./logger";
@@ -63,7 +64,10 @@ it("observes a payload-safe degradation transition separately from Run outcome",
 	};
 	const relay = createInMemoryLiveStreamRelay({
 		telemetry,
-		testHooks: { failEventPublishAtOrdinal: 0 },
+		testHooks: {
+			failEventPublishWhen: ({ eventType }) =>
+				eventType === EventType.RUN_STARTED,
+		},
 	});
 	const stream = await RunLiveStream.open({
 		relay,
@@ -127,7 +131,10 @@ it("observes a payload-safe degradation transition separately from Run outcome",
 it("retries a transient Live Stream failure-marker write after the Run terminalizes", async () => {
 	let markerAttempts = 0;
 	const relay = createInMemoryLiveStreamRelay({
-		testHooks: { failEventPublishAtOrdinal: 0 },
+		testHooks: {
+			failEventPublishWhen: ({ eventType }) =>
+				eventType === EventType.RUN_STARTED,
+		},
 	});
 	const stream = await RunLiveStream.open({
 		relay,
@@ -148,7 +155,13 @@ it("retries a transient Live Stream failure-marker write after the Run terminali
 });
 
 it.each([
-	["terminal publish", { failEventPublishAtOrdinal: 1 }, undefined],
+	[
+		"terminal publish",
+		{
+			failEventPublishWhen: ({ terminal }) => terminal,
+		} satisfies NonNullable<LiveStreamRelayOptions["testHooks"]>,
+		undefined,
+	],
 	["buffer overflow", undefined, { maxEvents: 1 }],
 ] as const)("latches %s failure while the Run still reaches its durable outcome", async (_name, testHooks, testLimits) => {
 	let markerAttempts = 0;
--- a/apps/agent-worker/src/run-loop.test.ts
+++ b/apps/agent-worker/src/run-loop.test.ts
@@ -392,7 +392,10 @@ describe("RunLoop — terminal outcomes", () => {
 
 	it("continues durable execution after a mid-text Live Stream write fails", async () => {
 		const liveStreamRelay = createInMemoryLiveStreamRelay({
-			testHooks: { failEventPublishAtOrdinal: 2 },
+			testHooks: {
+				failEventPublishWhen: ({ eventType }) =>
+					eventType === EventType.TEXT_MESSAGE_CONTENT,
+			},
 		});
 		const worker = buildWorker(1);
 		const loop = buildLoop(
@@ -435,7 +438,10 @@ describe("RunLoop — terminal outcomes", () => {
 
 	it("keeps committing Tool events after Live Stream publication fails", async () => {
 		const liveStreamRelay = createInMemoryLiveStreamRelay({
-			testHooks: { failEventPublishAtOrdinal: 1 },
+			testHooks: {
+				failEventPublishWhen: ({ eventType }) =>
+					eventType === EventType.TOOL_CALL_START,
+			},
 		});
 		const worker = buildWorker(1);
 		const loop = buildLoop(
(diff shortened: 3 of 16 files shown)


chengchao (commented):
Two-axis review (/code-review) of 0383b4b against #370 and ADR-0014. **Verdict: a genuinely rigorous suite — real components, non-vacuous assertions, no scope creep — but since this file IS the "advertise resumability" gate, three determinism/realism gaps should close before merge rather than after.**

## Spec

**Verified:** the producer is the production `RunLoop` + `Worker` (the synthetic worker is not imported; only the documented `RunProcessor` seam is injected); the consumer is the real chat-api app over HTTP, and the awaited `attach_attempt`/`retry` telemetry is recorded by the production route, not scaffolding; the production diff is empty (`apps/` and `packages/` untouched — `testLimits`/`testHooks` pre-exist on main). Assertions are exact — full SSE frame sequences, exact 503/410 bodies, `liveStreamFailedAt`, complete history message arrays — and hangs become failures via `AbortSignal.timeout` plus the suite timeout. Redis-unreachable and CI criteria fully pass.

**Gaps to close (each maps to a criterion this gate must prove):**

1. **"A Run attached while still queued streams from its first event after claim"** — the queued-state attach is not pinned. `app.request(...)` races `loop.tick()`; nothing gates the route's first attach attempt to happen before claim, so on an unlucky schedule the test degrades into the plain happy path. Force the ordering (e.g. hold the tick until the first attach attempt is observed via the existing telemetry hook).

2. **"Producer death mid-Run"** — death is `await producerRelay.close()`, a graceful close that DOES broadcast `relay_failed`; it is crash-equivalent only by ordering accident (the consumer subscribes after the broadcast is lost to retention-less pub/sub). That validates death-then-attach, but the **attached-then-death** variant — a reader already attached when the producer dies without any broadcast — is the silent-hang scenario this matrix exists to prove out, and it is untested. Kill the producer non-gracefully (drop the process/connection without close()) with a reader attached, and assert the reader converges via stale-Run recovery + history.

3. **"The terminal-instant attach race converges via the live terminal event or history"** — only the history arm is deterministically forced. The live-terminal arm is never exercised *in race* (scenario 1 covers it only without contention). Add the complementary forcing (release the backlog reply before the producer close instead of after).

Also noted: lease expiry via direct `UPDATE runs SET lockedUntil = epoch(0)` is an acceptable clock simulation, and the real `markStaleRunsTx` path is what terminalizes — good.

## Standards

No hard violations. Judgement items, in descending order of worth-fixing:

1. **`parseSse` drops the `id:`-line guard** its sibling `e2e/integration.test.ts` parser enforces — that guard is the executable form of ADR-0014's "no cursors on the wire". Restore it (or better, see next item).
2. **Fourth near-verbatim copy of the disposable-Redis harness** (`freePort` + `redis-server --bind 127.0.0.1 --save "" --appendonly no` + ping loop) across this file and the two live-text contract tests. Rule-of-three is past; extract a shared helper — same moment to share the SSE parser.
3. Minor: tautological `attachRetry.count() >= 1` immediately after `await attachRetry.observed`; `db.update(runs)` with no `.where()` (safe only while each test admits exactly one Run — fragile); `expect([200, 410]).toContain(...)` is a weak success criterion for a race a forced interleaving would make exact (falls out of Spec gap 3); the file runs in the `integration` CI job whose Postgres service it never uses, and has no env gate, so a local `bun test e2e/` without `redis-server` fails rather than skips.

Scope: exactly the test file + ci.yml, as the ticket demands. Frontier discipline held (#369 closed, PR #374 merged first).

chengchao (commented):
Round-two two-axis review of 20d25fc (amended). **Verdict: the gate did its job — the deeper attached-then-death test exposed a real production defect, and the fix is correct. Two items before merge: regression-lock the G2 test, and split or retitle the commit so the production fix isn't buried under `test:`.** Suites run in a scratch worktree: matrix 6/6, route 41/41, live-text 43/43, worker 31/31.

## Spec

**All three round-one gaps closed, mechanisms verified:**

- **G1 (queued attach)** — deterministic: the telemetry probe awaits `attach_attempt/retry` while asserting status is still `queued`, ticks the loop, and awaits `attach_attempt/success` before the gated processor releases the first event.
- **G2 (attached-then-death)** — realistic: producer relay connections are named (`testClientName`) and killed via `CLIENT KILL` with a reader already attached, so no broadcast is possible; stale-Run recovery (`markStaleRunsTx`) terminalizes; the attached stream closes incomplete with no synthesized terminal; reconnect gets 410; history shows `RUN_ERROR`.
- **G3 (terminal race)** — both arms forced: history arm via consumer pause across the terminal close → 410; live arm via producer `beforeBacklogReply` pause with `afterEventPublished` confirming the terminal published live before release.

**The production change is defect-found-by-gate, not scope creep.** The attached-then-death test exposed that a connected reader of a silently-dead producer had no route-side give-up condition — keepalive failure and client abort never fire in that state, so the reader hung until its own timeout. `watchRunForLiveStreamRecovery` fixes it compliantly: Postgres-governed polling stays in the route (per ADR-0014's "the caller owns retry"), bounded (5s interval, single-in-flight, cleared on every exit path in both attach paths), and it aborts to an incomplete close — no fabricated terminal, no wire ids. The relay hunks (`failEventPublishWhen` predicate replacing magic ordinals, `testClientName` connection naming) are inert test seams on the existing `testHooks` surface. No #369 regressions: fail-stop latch, terminal-after-commit, 503/410, and abort-on-give-up all intact.

**Before merge:**

1. **Regression-lock G2.** Verified empirically: with the round-one route reverted, the G2 test still passes (13.1s, rescued by its own `AbortSignal.timeout(10_000)`) versus 8.0s with the fix — i.e. the suite currently does not fail on the exact hang the production fix prevents. Drop that timeout signal (let the suite timeout be the failure) or assert convergence latency well below it.

2. Benign race, note only: the recovery watch reads terminal status, which commits before the terminal live publish; a poll landing in that window can abort a healthy reader just before `RUN_FINISHED`, converging via 410/history instead. Correct per contract — worth a one-line comment at the poll site.

## Standards

**All round-one items resolved:** `e2e/ag-ui-sse.ts` restores the `id:`-line guard and is shared by the e2e and matrix suites; `test-support/redis-test-server.ts` genuinely replaces all four hand-rolled redis spawns; the tautological probe is gone; the unscoped update is now scoped; CI moved to the ungated `test` job (which installs redis-server), eliminating the unused-Postgres-service placement.

**New items, in descending order:**

1. **Commit hygiene (fix before merge):** a single `test:`-titled commit now ships a production behavior fix. Split into `fix(chat-api): end attached readers on silent producer death` + `test: verify ADR-0014 failure matrix (#370)` — or at minimum retitle — so the defect fix is findable in history.
2. Judgement calls: `streamAgUiRun` vs `waitForAndStreamAgUiRun` duplication slightly worsened (watcher setup threaded independently in both; six `stopRecoveryWatch` exit points); the raw-SQL scoped update in the e2e file deserves a one-line comment naming the single-drizzle-instance constraint that forced it; recovery polling reuses `LIVE_STREAM_KEEPALIVE_MS` where a dedicated constant would name the concept; top-level `test-support/` is a new convention (a workspace package would be cleaner, prior art: `@mymemo/agent-db`'s exported testing module); `scripts/smoke/agent-conversation-smoke.ts` still captures `id:` rather than rejecting it — pre-ADR-0014 legacy worth a follow-up.

chengchao (commented):
Round-three review of 7e678ed. **Verdict: mergeable — both round-two asks are resolved, the last one with mutation-grade evidence.**

## Spec — G2 regression lock proven empirically

- **Baseline** (7e678ed, redis-server 6.2.11): matrix suite 6/6 in ~8.3s; the producer-death test runs 5.06s with no client-side AbortSignal (the round-three change), while the other reconnect tests keep their short timeouts.
- **Mutation** (`watchRunForLiveStreamRecovery` neutralized to a no-op, tests untouched): the producer-death test **fails via the 20s test-level timeout**, and the hung SSE reader cascades timeouts into the three subsequent tests — exactly the silent-hang signature this gate exists to catch. 2 pass / 4 fail, one dangling process killed.
- **Restore**: 6/6 again (~8.1s); no flakiness across three suite runs.

The suite now genuinely fails without the production fix it forced into existence. That closes the round-two blocker.

## Standards — settled by inspection

The 10-line delta contains precisely the asked-for items and nothing else: `LIVE_STREAM_RECOVERY_POLL_MS` is a named constant; the benign terminal-race is documented at the poll site ("Terminal state may commit just before its live terminal publish..."); the raw-SQL update carries the single-drizzle-instance rationale; and the commit is retitled `fix(chat-api): end attached readers on silent producer death (#370)` — the accepted retitle option, so the production fix now leads the history entry.

## Tracked follow-ups (non-blocking, for the record)

- `scripts/smoke/agent-conversation-smoke.ts` still captures `id:` lines rather than rejecting them (pre-ADR-0014 legacy).
- `streamAgUiRun` / `waitForAndStreamAgUiRun` watcher-setup duplication — candidate for consolidation next time the route is touched.
- `test-support/` as a proper workspace package (prior art: `@mymemo/agent-db`'s exported testing module).
- Store-era telemetry vocabulary pruning rides with #371.

With this merged and #370 closed, the frontier is #371 — delete the retained store — and the ADR-0014 chain is complete.

PR #409, opened 2026-08-01
feat(agent-db): fence Run writes on ownership epoch

chengchao:
## Summary

- fence Run event appends, terminal transitions, and active Live Stream failure markers on the Conversation Ownership epoch
- return the shared typed fence-rejection vocabulary from append and marker paths while preserving append-class and atomic Tool-batch semantics
- make the worker apply uniform halt, skip, and interruption decisions across every fenced Run write
- authorize post-terminal failure markers by immutable terminal status instead of legacy Run lease columns

## Why

Run-state writes still depended on the transitional Run lease after execution ownership moved to the Conversation. A superseded worker therefore needed the Ownership epoch enforced at every Run-state mutation, including the sequence-number allocation itself.

## Impact

This is an internal ownership-safety change. Permanent Conversation history, its projection, and AG-UI event shapes remain unchanged.

## Validation

- `env -u E2B_API_KEY bun run test`
- TypeScript checks for `packages/agent-db`, `apps/agent-worker`, and `apps/chat-api`
- Biome checks on all changed files
- two-axis standards and issue-spec review, with no unresolved findings

The environment-gated real-PostgreSQL concurrency cases remain wired to the integration lane and were skipped locally because `AGENT_DATABASE_URL` was not configured.

Closes #399


chengchao (commented):
## Two-axis review — `main`@6528f72 ... `codex/issue-399-epoch-run-fences`

Diff: `git diff main...pr-409` (9 files, +771/−455). Spec: #399.
The two axes are reported separately and deliberately not merged or reranked.

---

## Standards

**No hard violations found.** Biome/tsc-enforced items skipped.

- **Doc drift (minor, `AGENTS.md` module table)** — the diff moves appends, terminal transitions and the failure marker onto the epoch, but `AGENTS.md` still describes `src/run-ownership.ts` as supplying the fence "additionally user identity for Conversation runtime **and terminal state**", and `run-store.ts` as the "lease"-fenced helper set. Neither `AGENTS.md` nor `CONTEXT.md` is touched.
- `CONTEXT.md` "_Avoid_: run lease" is **not** breached — the diff consistently qualifies it as the *legacy Run lease*, a real distinct mechanism.

Baseline smells (all judgement calls):

1. **Duplicated Code — `packages/agent-db/src/run-store.ts` + `packages/agent-db/src/artifact-store.ts`.** The same catch-and-reclassify shape appears twice:

   ```ts
   if (!(error instanceof TerminalFenceChangedError)) throw error;
   return { outcome: "rejected", ...(await classifyRunWriteRejectionTx(db, input.owner)) };
   ```

   It is the only reason `TerminalFenceChangedError` and `classifyRunWriteRejectionTx` are exported at all (the comment concedes this: "exported only so the artifact composer can preserve the same transaction boundary"). A `runTerminalTransaction(db, owner, fn)` wrapper would own the boundary once — and would collapse smell 4.
2. **Data Clumps — `apps/agent-worker/src/run-loop.ts`.** `drain` is threaded as a new parameter through `finish`, `failRun`, `terminalize`, `publishArtifactsAndFinish`, `reconcileRejectedTerminal`, and `(owner, drain, entry)` through both append methods — six signatures widened solely to reach `noteRejectedRunWrite`. `{owner, drain, entry}` is a served-Run context wanting one type.
3. **Repeated type shape — `run-store.ts`.** `({ outcome: "rejected" } & FenceRejection)` is now spelled four times (`TerminalTransitionResult`, `AppendRunEventResult`, `AppendRunEventsResult`, `MarkLiveStreamFailedResult`), then re-derived in `run-loop.ts` as `RejectedRunWrite` and via `Extract<…>` as `RejectedTerminal`. One exported `RunWriteRejected` would serve all six. The `seq?: never` padding on the rejected arm is a symptom.
4. **Surgical Changes pressure — `artifact-store.ts`.** 183 changed lines for a semantic delta of ~10; the rest is re-indentation from the `try` wrapper. Extracting the tx body (or fix 1) keeps the real change reviewable.
5. **Flag-ish optional param — `noteRejectedRunWrite(drain, runId, rejection, entry?)`.** The optional trailing `entry` switches the method between "log and skip" and "mutate Run state + abort three controllers", with a nested cascade on `rejection.rejected` / `rejection.current`; `runId` is also redundant whenever `entry` is present. Two behaviours in one method.
6. **Empty branch — `serveRun` catch.** `if (error instanceof RunWriteRejectedError) { /* comment */ } else if …` — control-flow-by-exception whose handler is a comment.

---

## Spec

**Verdict: all nine acceptance criteria are met.** Tests (130) pass; all three workspaces typecheck.

### (a) Missing/partial

Nothing missing. One weakening worth calling out against *"The classify read that names a rejection moves onto the epoch together with the writes it explains."* The read does move onto the epoch (`classifyRunWriteRejectionInTx` → `conversationEpochExists`), but for the terminal path it is no longer *in* the rejecting transaction: `commitLockedRunTerminalInTx` raises the new `TerminalFenceChangedError`, the transaction rolls back, and `transitionRunTerminalTx` / `publishArtifactsAndTransitionRunDoneTx` classify in a *fresh* transaction (`packages/agent-db/src/run-store.ts:659-668`, `packages/agent-db/src/artifact-store.ts:176-183`). The module doc above `classifyFenceRejectionInTx` still claims "One read, in the rejecting transaction." It is safe only because a lapsed lease cannot be revived (`renewConversationLeaseTx` requires `owner_until > now()`) and a re-Claim bumps the epoch — so `fenceHolds` cannot flip back to true and be misreported as a skippable `status` refusal. That invariant is load-bearing and undocumented at the new catch sites.

### (b) Scope creep

Spec: *"this ticket only changes which predicate the fence evaluates and widens the typed vocabulary to the remaining paths."* Three behaviours exceed that:

1. `RunWriteRejectedError` + `RunEndState.skipTerminalization`, and `noteRejectedRunWrite` now *aborting the in-flight processor* (`entry.controller` / `interruptionController` / `ownershipLostController`) — `apps/agent-worker/src/run-loop.ts:660-686`. The spec asked only that appends *return* the rejection so halt-vs-skip is uniform; actively cancelling the turn and suppressing terminalization is new drain control flow.
2. `reconcileRejectedTerminal` now feeds its unresolved rejection into `noteRejectedRunWrite` (`run-loop.ts:1081`), so a rejected *terminal* transition sets `drain.halted` and suppresses release. Previously terminal rejections did not affect the drain. Correct-looking, but not requested.
3. New early return in `executeServedRun` after `RunLiveStream.open` (`run-loop.ts:800-803`).

### (c) Implemented but looks wrong

Nothing rises to a defect. Two notes: the marker callback now throws a bare `Error` purely as control flow after the rejection was already recorded (`RunLiveStream.#persistFailureMarker` swallows it), so the throw carries no information; and the marker's terminal branch is now `inArray(runs.status, TERMINAL_RUN_STATUSES)` with no epoch conjunct — any caller matching user/conversation/run may set it. That is exactly what the criterion asked for ("identifies an immutable Run by its terminal status"), so it is intended widening, not a bug.

---

**Summary:** Standards — 7 findings (0 hard, 6 judgement-call smells + 1 doc drift); worst is the duplicated catch-and-reclassify block forcing `TerminalFenceChangedError` / `classifyRunWriteRejectionTx` to be exported. Spec — 5 findings (0 missing requirements); worst is the terminal path's classify read moving out of the rejecting transaction, correct only via a load-bearing, undocumented invariant that a lapsed lease can never be revived.

PR #410, opened 2026-08-01
feat(agent-db): reclaim lapsed Conversations

chengchao:
## Summary

- reclaim lapsed Conversation Ownership leases with a Conversation-row `FOR UPDATE SKIP LOCKED` transaction
- terminalize started `running` and `interrupt_requested` Runs, taint the Workspace when command cleanup is unproven, and clear Ownership atomically
- preserve never-started queued Runs so the released Conversation is immediately eligible for its next Claim
- preserve interruption, `stale_worker`, and Live Stream degradation semantics
- retain the old unowned queue-age timeout as a separate, Run-index-driven liveness-sweep path
- add PGlite, worker-loop, scaler, and real-Postgres concurrency coverage

## Why

A worker can disappear while holding Conversation Ownership. The prior Run-scoped cleanup path could not atomically finalize the Conversation's started Runs, prevent unsafe Workspace reuse, and release the Conversation for a fresh Claim.

A lapsed lease is deliberately not directly Claimable: Reclamation first establishes the durable boundary, then clears Ownership. This prevents a fresh holder from reconnecting to or writing alongside the vanished holder. Never-started queued Runs survive that boundary and are served by the next Claim.

This makes the Conversation the Reclamation unit, matching ADR-0015 and the Claim protocol.

Closes #400.

## Validation

- focused Reclamation, queue, worker-loop, scaler, event, and ownership suites — 176 passed
- real PostgreSQL 17 contention suite — 10 passed, covering concurrent Reclamation and Claiming, `SKIP LOCKED`, Claim/admission snapshot boundaries, and superseded-holder fencing
- `bunx tsc --noEmit -p packages/agent-db/tsconfig.json`
- `bunx tsc --noEmit -p apps/agent-worker/tsconfig.json`
- `bunx tsc --noEmit -p apps/worker-scaler/tsconfig.json`
- `bunx tsc --noEmit -p apps/chat-api/tsconfig.json`
- Biome checks passed
- previous repository-wide run reached 496 passing worker tests; one timing-sensitive SDK supervision test timed out in the full run and passed independently in 161 ms


chengchao (commented):
Two-axis review (`/code-review` vs issue #400): **Spec** (acceptance criteria) and **Standards** (AGENTS.md / CONTEXT.md / ADRs + smell baseline). Details are in the inline comments; summary:

**Spec — 5 findings.** Worst: Reclamation terminalizes never-started `queued` Runs as `error` — `ACTIVE_RUN_STATUSES` includes `queued`, but a Run queued moments before the owner vanished isn't the spec's "nobody picked this up" signal; on `main` a successor Claim served it. Also: the Claim-predicate narrowing and the taint-on-every-lapse go beyond what the ACs asked (both defensible, both worth an explicit decision), the queue-age sweep seq-scans `conversations`, and the legacy index survives (deliberate, deferred to #402).

**Standards — no hard violations.** Findings are judgement calls: naming drift (`claimableConversations`, `reclamationOutcome`), `OWNERSHIP_LEASE_MS` doing double duty as the queue-age threshold, two shapes for the same sweep loop in `run-loop.ts`, one dropped AGENTS.md clause, and one locking question on the un-skip-locked run-row `FOR UPDATE`.

The three highest-severity claims (Claim predicate, `queued` in the reclaim selection, the run-row lock) were verified against the diff, not just pattern-matched.

packages/agent-db/src/run-store.ts
```diff
@@ -1155,69 +1174,162 @@ export async function markStaleRunsTx(
...
-					and ${runs.lockedUntil} <= now()
-				) or (
-					${runs.status} = 'queued'
-					and ${runs.createdAt} <= now() - interval '${sql.raw(
-						String(LOCK_DURATION_MS),
-					)} milliseconds'
+				and(
+					eq(runs.userId, candidate.userId),
+					eq(runs.conversationId, candidate.conversationId),
+					inArray(runs.status, ACTIVE_RUN_STATUSES),
```
chengchao: **[spec — correctness]** `ACTIVE_RUN_STATUSES` includes `queued`, so Reclamation terminalizes never-started queued Runs as `error`. The spec draws the line differently: *"The queue-age rule — a Run queued longer than the lease duration fails — fires only when its Conversation is unowned, which is the actual 'nobody picked this up' signal."* A Run queued seconds before the owner vanished (crash between Claim and start, or mid-drain arrival) hasn't hit that signal — on `main` a successor Claim served it.

Since this transaction clears the ownership columns anyway, excluding `queued` here (terminalize only `running`/`interrupt_requested`) would let the freed Conversation be re-Claimed to serve those Runs, while `expireUnownedQueuedRunsTx` still catches genuinely abandoned ones. It would also make the *"immediately claimable again"* AC demonstrable without seeding a fresh Run — as the tests currently must.

packages/agent-db/src/run-store.ts
```diff
@@ -1155,69 +1174,162 @@ export async function markStaleRunsTx(
...
+				),
+			)
+			.for("update");
```
chengchao, line 1196: **[standards — locking, worth verifying]** This `.for("update")` on `runs` has no `skipLocked`, and it's taken while holding the Conversation row. If the lapsed owner is alive-but-slow rather than dead — e.g. a long transaction still holding one of these run rows (the artifact commit's fence-first Run lock is the obvious candidate) — Reclamation waits here *while holding the Conversation row*, which `admitQueuedRunTx` takes blockingly. The new "concurrent reclaimers split the work instead of blocking" comment only covers the candidate select above.

Locks die with the owner's connection, so the common crash case is fine — but the wait is unbounded in the slow-owner case. Either a `nowait`/skip-and-retry strategy, or a comment arguing why the hold time is bounded (epoch fences make the old owner's write reject, but rejection doesn't release a lock mid-transaction).

packages/agent-db/src/run-store.ts
```diff
@@ -1155,69 +1174,162 @@ export async function markStaleRunsTx(
...
+			tx,
+			runsToReclaim,
+		);
+		const reclaimedRuns = terminalizedRuns.map(
+			({ liveStreamFailureMarkedNow, ...run }) => ({
+				...run,
+				liveStreamFailureMarkedByReclamation: liveStreamFailureMarkedNow,
+			}),
+		);
+		await taintRuntimeSandboxForReclamationInTx(tx, candidate);
```
chengchao: **[spec — scope]** The AC carries the #391 taint *"into that same transaction, now keyed on the Conversation the sweep already holds"* — a re-keying, not a widening. This taints on **every** lapsed lease, including one with zero Active Runs (worker crashed after the Outcome committed but before release — nothing was mid-command, so cleanup isn't unproven). Cost: the next turn provisions a fresh Workspace and the paused-sandbox files (the ADR-0007 persistence story) are discarded for an ordinary crash-after-done.

Consider tainting only when `runsToReclaim` contained a `running`/`interrupt_requested` Run — i.e. exactly when command cleanup genuinely can't be proven, matching the `taint` column's own doc comment.

packages/agent-db/src/run-store.ts
```diff
@@ -1155,69 +1174,162 @@ export async function markStaleRunsTx(
...
+		};
+	});
+}
+
+/**
+ * Expire old queued Runs only when their Conversation is unowned. The
+ * Conversation row is locked with `FOR UPDATE SKIP LOCKED`, so this backstop
+ * cannot race admission or Claim and never waits behind their lifecycle lock.
+ */
+export async function expireUnownedQueuedRunsTx(
```
chengchao, line 1251: **[perf]** This candidate select is `conversations WHERE owner_until IS NULL AND EXISTS(queued run older than lease)`. Unowned is the steady state of every Conversation, and `conversations_reclamation_idx` is partial on `owner_until IS NOT NULL` — so this is a full scan of the unboundedly growing `conversations` table plus a correlated EXISTS, per worker, every tick. That's the exact pattern `queue-metrics.ts` documents the hot paths as avoiding (*"neither scans the unboundedly growing `conversations` table"*).

Drive the candidate from `runs_queue_claim_idx` (status `queued`, `created_at <= now() - lease`) and probe/lock `conversations` by PK, the way the Claim does.

**[standards — duplication, judgement call]** Also: this function and `reclaimConversationTx` share the whole candidate-select → runs-select → `terminalizeRunsForLivenessSweepInTx` → `{userId, conversationId, runs}` shape; only the innermost loop was extracted.

packages/agent-db/src/conversation-ownership.ts
```diff
@@ -114,7 +117,7 @@ export async function claimConversationTx(
 					  from ${conversations} c
 					  join ${runs} r using (user_id, conversation_id)
 					 where r.status = 'queued'
-					   and (c.owner_until is null or c.owner_until <= now())
+					   and c.owner_until is null
```
chengchao, line 120: **[spec — scope creep, decide explicitly]** No AC asks for this narrowing — AC1 only requires Reclamation to take *"the same row lock a Claim takes"*. Dropping `owner_until <= now()` from the Claim predicate is defensible (a fresh Claim would otherwise absorb a lapse without terminalizing the vanished owner's Runs or tainting its Workspace), but it creates a new failure mode: a lapsed Conversation is unservable until some worker's Reclamation sweep succeeds, where previously a fresh Claim alone unblocked it. Reclamation is now on the serving critical path — if that's the intent (the doc comment above suggests it is), it deserves a line in the PR description / ADR-0015, and the queue metric's name should follow (see comment in `queue-metrics.ts`).

packages/agent-db/src/conversation-ownership.ts
```diff
@@ -32,7 +32,7 @@ import { conversations, runs } from "./schema";
  * the Run lease uses, renewed on the worker's 15s heartbeat — four missed
  * renewals before the Conversation becomes reclaimable.
  */
-const OWNERSHIP_LEASE_MS = 60_000;
+export const OWNERSHIP_LEASE_MS = 60_000;
```
chengchao: **[standards — smell, judgement call]** Exported solely so `expireUnownedQueuedRunsTx` can reuse a *lease* duration as a *queue-age* threshold. Two unrelated policies now move together — tuning the heartbeat/lease silently retunes how long a queued Run may wait before the backstop kills it. Give the backstop its own named constant (fine for it to equal the lease today), so the coupling is a visible choice rather than an accident.

packages/agent-db/src/queue-metrics.ts
```diff
@@ -12,9 +12,10 @@ export interface ConversationQueueMetrics {
  * Conversations rather than Runs: under Conversation ownership (ADR-0015) the
  * claimable unit is the Conversation — several Active Runs on one Conversation
  * are one drain — so a Run-shaped metric would request tasks that cannot
- * possibly help. Claimable means at least one recent queued Run and no live
- * Ownership lease; owned means a live lease with a recent Active Run —
- * a Conversation occupying a worker slot for work inside the window.
+ * possibly help. Claimable demand means at least one recent queued Run and no
+ * live Ownership lease: an unowned Conversation can be Claimed, while a lapsed
+ * one first needs Reclamation. Owned means a live lease with a recent Active Run
```
chengchao: **[standards — naming]** With the Claim predicate narrowed to `owner_until IS NULL`, the SQL below still counts `owner_until <= now()` rows, so `claimableConversations` now includes Conversations that are **not** claimable until Reclamation clears them. The comment explains the intent (lapsed = demand, so a worker boots to reclaim), but the field name — and anything downstream reading "claimable" — now disagrees with `claimConversationTx`. Rename (`demandConversations`?) or split lapsed leases into their own count so the scaler can tell backlog from crash debt.

packages/agent-db/src/run-events.ts
```diff
@@ -216,7 +216,7 @@ export function validateDurableRunEventSequence(
 	>();
 	let terminalSeen = false;
 	let terminalOutcome: RunOutcome | null = null;
-	let staleRecoveryOutcome = false;
+	let reclamationOutcome = false;
```
chengchao: **[standards — naming, judgement call]** `reclamationOutcome` is set for *any* terminal with `reason === "stale_worker"`, which per the new `ExpiredQueuedRuns` doc comment includes the non-Reclamation path (*"This is the retained queue-age backstop, not Reclamation: no Ownership lease was lost"*). The rename asserts a distinction the value doesn't hold. Either name it for what it actually keys on (`crashedOwnerOutcome` / `livenessSweepOutcome`), or give the two sweeps distinct reasons — the spec leans on the reason as *"the durable proof that an incomplete Tool prefix came from a crashed owner"*, and the queue-age path has no such prefix.

apps/agent-worker/src/run-loop.ts
```diff
@@ -373,53 +375,90 @@ export class RunLoop {
...
 		}
 	}
 
-	private async tryRecoverStaleRuns(): Promise<void> {
+	private async tryRunLivenessSweep(): Promise<void> {
+		await this.tryReclaimConversations();
+		await this.tryExpireUnownedQueuedRuns();
+	}
+
+	private async tryReclaimConversations(): Promise<void> {
```
chengchao, line 398: **[standards — smell, judgement call]** Two shapes for the same drain-until-null sweep: `tryReclaimConversations` is a pure try/catch Middle Man around `reclaimConversations` (which holds the loop), while `tryExpireUnownedQueuedRuns` inlines its loop inside the `try`. Pick one shape — a small shared `drainSweep(label, sweepTx, onResult)` would remove both the asymmetry and the duplicated `for(;;) { … if (!x) break; logger.warn({…runs.map(…)}) }` skeleton.

AGENTS.md
```diff
@@ -72,8 +72,8 @@ Source: [forrestchang/andrej-karpathy-skills](https://github.com/forrestchang/an
...
...d `conversation_runtime`/`orphan_sandboxes` helpers, the SDK session-transcript mirror helpers over the worker-only `agent_sessions` table, the canonical `run_events.type` write-side vocabulary (`RunEventType`), the PGlite test harness, and the migrations. The Claim and Reclamation protocol's concurrency properties — `SKIP LOCKED` contention, the Claim-versus-admission snapshot boundary, concurrent reclaimers, and a superseded holder racing its successor — are the one thing a single in-process backend cannot express, so they live in `src/conversation-ownership.postgres.test.ts`, gated on `AGENT_DATABASE_URL` and run by the CI `integration` job against the Postgres major production runs. Both runtimes import it so the queue protocol, runtime-persistence fence, and permanent-history event vocabulary share one definition over one `pg` driver. Runtime-store performs sandbox/taint and Agent-session pointer writes; run-store composes its named operations — including atomic Conversation Reclamation that terminalizes Active Runs, taints the Workspace, and clears Ownership — while chat-api's `PostgresRunStore` composes transaction-scoped shared admission with the Conversation lifecycle lock
```
chengchao: **[docs]** This rewrite drops "the pointer publication into qualifying terminal transactions through its transitional user-bound Run lease" from the run-store composition list, but `publishAgentSessionPointerInTx` is untouched by this PR and still composed into qualifying terminal transactions. That clause documents a live behavior (its lease fence is only removed in #401/#402) — restore it alongside the new Reclamation sentence.

packages/agent-db/src/schema.ts
```diff
@@ -364,7 +364,8 @@ export const runs = pgTable(
 		index("runs_queue_claim_idx")
 			.on(t.createdAt)
 			.where(sql`${t.status} = 'queued'`),
-		// Stale-run recovery: active runs whose lock deadline has passed.
+		// Legacy Run-lease access path, removed with the lease in #402. Reclamation
+		// scans `conversations.owner_until`; no production path scans this index.
 		index("runs_stale_recovery_idx")
```
chengchao, line 373: **nit:** Keeping the index annotated rather than dropping it mid-transition is the right call — but nothing besides this comment tracks the removal. Worth confirming #402's scope explicitly includes the migration dropping `runs_stale_recovery_idx` (and eventually the lease columns it serves).

packages/agent-db/src/run-store.ts
chengchao: Fixed in d17aa04. Reclamation now selects only `running` and `interrupt_requested` Runs. Never-started queued Runs remain queued, and the tests now prove the released Conversation's next Claim receives them.

packages/agent-db/src/run-store.ts
chengchao: Kept the blocking Run-row lock intentionally and documented the invariant in d17aa04. Skipping a locked started Run while clearing Conversation Ownership would strand it. Any holder acquired the row before the lease lapsed; these transactions are database-only and short, and Postgres rechecks the status predicate after the holder commits.

packages/agent-db/src/run-store.ts
chengchao: Fixed in d17aa04. The Reclamation transaction taints only when it found a started Active Run; a lapsed lease with no such Run still clears Ownership but preserves the Workspace. Added a PGlite regression test.

packages/agent-db/src/run-store.ts
chengchao: Fixed the performance issue in d17aa04: queue-expiration candidates now drive from `runs_queue_claim_idx`, join/probe the Conversation by key, order by queued `created_at`, and lock only the Conversation with `FOR UPDATE OF c SKIP LOCKED`. I kept the two public transactions separate because Reclamation and queue expiration have different domain rules; they share only the terminalization primitive.

packages/agent-db/src/conversation-ownership.ts
chengchao: Keeping the narrowed Claim predicate intentionally. d17aa04 records the decision in ADR-0015 and the PR description: a lapsed lease must pass through Reclamation before a fresh Claim can reconnect or write, while queued Runs survive and become immediately Claimable after Ownership is cleared.

packages/agent-db/src/conversation-ownership.ts
chengchao: Fixed in d17aa04. Queue expiration now has its own `UNOWNED_QUEUE_TIMEOUT_MS` policy constant; it equals the current 60-second lease explicitly rather than importing/exporting the lease constant.

packages/agent-db/src/queue-metrics.ts
chengchao: Fixed in d17aa04. The metric/API is now `demandConversations`, with matching SQL aliases, scaler usage, tests, and documentation. It intentionally includes both directly Claimable work and lapsed Conversations that need Reclamation.

packages/agent-db/src/run-events.ts
chengchao: Fixed in d17aa04. Renamed the sequence flag to `livenessSweepOutcome`, matching the shared persisted `stale_worker` reason used by both Reclamation and the retained queue-age path.

apps/agent-worker/src/run-loop.ts
chengchao: Fixed in d17aa04 by using the same inline try/catch + drain-until-null shape for both sweeps and removing the single-purpose `reclaimConversations` middle man. I avoided adding a generic callback abstraction because the result handling differs materially.

AGENTS.md
chengchao: Fixed in d17aa04. Restored the Agent-session pointer-publication clause alongside the new Reclamation description.

packages/agent-db/src/schema.ts
chengchao: Confirmed: issue #402's first acceptance criterion explicitly drops `runs_stale_recovery_idx` together with the Run lease columns. The existing schema comment already points to #402, so no additional code change is needed in this PR.

PR #426, opened 2026-08-05
feat(agent-worker): validate Vega-Lite chart payloads

chengchao:
## Summary

- add `chart` to the PresentUI catalog validator
- validate specs against the vendored Vega-Lite 5.23.0 JSON Schema with AJV, compiled lazily once per worker process
- enforce the exact 8 KiB spec cap, inline array data only, and a 200-row aggregate cap across nested specs and lookup transforms
- return stable repair rule codes for schema, size, data-source, data-shape, and row-count violations
- cover the prototype fixture, nesting positions, exact boundaries, and compilation caching

## Motivation and impact

Chart payloads are model-authored and therefore untrusted. This keeps the server-side contract deterministic and display-only: only schema-valid, bounded, self-contained Vega-Lite specs are admitted, while URL, named, generated, and non-countable data sources are rejected before persistence.

The worker does not install or execute Vega/Vega-Lite; rendering and runtime hardening remain client-owned.

## Validation

- `bun run test` — 931 passed, 13 intentionally skipped Postgres concurrency tests
- `bun test apps/agent-worker/src/ui-payload-validator.test.ts --timeout=30000` — 44 passed
- `bunx tsc -p apps/agent-worker/tsconfig.json`
- `bunx biome check apps/agent-worker/src/ui-payload-validator.ts apps/agent-worker/src/ui-payload-validator.test.ts apps/agent-worker/package.json`
- two-axis standards/spec review — no remaining findings

Closes #420


apps/agent-worker/src/ui-payload-validator.test.ts
```diff
@@ -64,8 +65,70 @@ function tableAtSerializedEventBytes(targetBytes: number) {
...
+				},
+				color: { value: "#4f46e5" },
+			},
+		},
+	},
+};
+
 describe("validateUiPayload", () => {
+	it("compiles the pinned Vega-Lite schema lazily once", () => {
+		const compile = spyOn(Ajv.prototype, "compile");
```
chengchao, line 117: **Standards (judgement call): order-dependent test.** This asserts exactly one `Ajv.prototype.compile` call, which only holds because this test happens to be the first chart validation to touch the module-level `cachedVegaLiteValidator`. Reordering tests in this file — or any other test file in the same process validating a chart first — breaks it. Consider asserting the *delta* (compile count doesn't increase between the first and second validation) instead of an absolute count, so the invariant survives test reordering.

apps/agent-worker/src/ui-payload-validator.ts
```diff
@@ -94,6 +114,9 @@ const COMPONENT_DEFINITIONS: Record<SupportedComponent, ComponentDefinition> = {
 
 const TABLE_ALIGNS = ["left", "center", "right"] as const;
 const CARD_TONES = ["neutral", "info", "success", "warning", "danger"] as const;
+const CHART_SCHEMA_DETAIL_CHARACTERS = 300;
```
chengchao: **Standards (judgement call): duplicated magic number.** `CHART_SCHEMA_DETAIL_CHARACTERS` is unexported, so the test re-hardcodes `toBeLessThanOrEqual(300)` (ui-payload-validator.test.ts:244). Exporting it (or referencing it from the test) keeps the two from drifting. Keeping it out of `UI_PAYLOAD_LIMITS` looks right — it isn't an ADR-0017 wire cap.

apps/agent-worker/src/ui-payload-validator.ts
```diff
@@ -165,6 +188,144 @@ function serializedEnvelopeBytes(payload: Record<string, unknown>): number {
...
+	nonInlinePath?: string;
+	invalidValuesPath?: string;
+} {
+	let rows = 0;
+	let nonInlinePath: string | undefined;
+	let invalidValuesPath: string | undefined;
+
+	const visitData = (data: unknown, path: string) => {
+		if (nonInlinePath !== undefined || invalidValuesPath !== undefined) return;
+		if (!isRecord(data) || !Object.hasOwn(data, "values")) {
```
chengchao, line 273: **Spec (minor over-strictness):** `data: null` is schema-valid Vega-Lite for "no data", but this branch classifies it as `chart_data_not_inline` with the "must use inline data.values" repair hint. The issue says non-inline data *sources* are rejected — null is no source — so the repair hint points the model's repair loop the wrong way for a spec that may just not need data. Either admit `data: null` or give it a detail that doesn't instruct adding `values`.

apps/agent-worker/src/ui-payload-validator.ts
```diff
@@ -165,6 +188,144 @@ function serializedEnvelopeBytes(payload: Record<string, unknown>): number {
...
+
+	const visitSpec = (candidate: unknown, path: string) => {
+		if (
+			!isRecord(candidate) ||
+			nonInlinePath !== undefined ||
+			invalidValuesPath !== undefined
+		) {
+			return;
+		}
+		if (Object.hasOwn(candidate, "datasets")) {
```
chengchao, line 292: **Spec (scope note):** wholesale rejection of `datasets` at any nesting level — even purely inline ones — isn't in issue #420, which enumerates data-block positions but never `datasets`. It's coherent (named-data references would be rejected anyway, and allowing inline `datasets` would complicate the row-cap sum), but worth confirming it's intended rather than accidental.

apps/agent-worker/src/ui-payload-validator.ts
```diff
@@ -165,6 +188,144 @@ function serializedEnvelopeBytes(payload: Record<string, unknown>): number {
...
+				visitSpec(child, `${path}.${property}[${index}]`);
+			}
+		}
+		if (Object.hasOwn(candidate, "spec")) {
+			visitSpec(candidate.spec, `${path}.spec`);
+		}
+	};
+
+	visitSpec(spec, "chart spec");
+	if (nonInlinePath !== undefined) return { rows, nonInlinePath };
```
chengchao, line 326: **Standards (judgement call): duplicated ceremony in `inspectChartData`.** The first-error short-circuit `nonInlinePath !== undefined || invalidValuesPath !== undefined` appears in both `visitData` and `visitSpec`, and this three-branch tail is behaviorally identical to a single `return { rows, nonInlinePath, invalidValuesPath }`. A small discriminated-union result (matching this file's own `UiPayloadValidationResult` idiom) would remove both.

apps/agent-worker/src/ui-payload-validator.test.ts
```diff
@@ -64,8 +65,70 @@ function tableAtSerializedEventBytes(targetBytes: number) {
...
+	const spec = {
+		data: { values: [{ value: 1 }] },
+		mark: "point",
+		description: "",
+	};
+	const baseBytes = new TextEncoder().encode(JSON.stringify(spec)).byteLength;
+	return { ...spec, description: "x".repeat(targetBytes - baseBytes) };
+}
+
+const prototypeChart = {
```
chengchao: **Standards (weak judgement call): naming.** `prototypeChart` doesn't say what the fixture *is* — a known-valid exemplar. Something like `validChart` (or `exemplarChart`) would read more honestly at the call sites that use it as the generic happy-path payload.

chengchao (commented):
Two-axis review (Standards / Spec vs issue #420). All six acceptance criteria are implemented and tested — exemplar fixture matches the prototype field-for-field, `data.url` is rejected at every nesting position, the row cap sums across data blocks with exact boundaries, the 8 KiB boundary is exact, and lazy one-time compilation is verified. No hard standards violations; rule codes, detail-message shape, and the `ComponentDefinition` entry all follow the file's existing idioms.

Findings worth a look are inline. Two file-level notes that don't fit a line comment:

1. **Schema pin is unverifiable from the repo** — the vendored `vega-lite-v5.23.0.schema.json` carries no internal version metadata (upstream's build artifact genuinely has no `$id`), so the 5.23.0 pin rests solely on the filename and a code comment. Structure sanity-checks out (draft-07, `$ref: TopLevelSpec`). Consider recording the upstream source URL + a content hash in the provenance comment so the pin can be re-verified.
2. **`chart_data_values_invalid`** is a fifth rule code the issue didn't enumerate (non-array inline `values`, e.g. Vega-Lite's CSV-string form). It's needed to keep the 200-row cap countable and it fails closed with tests — flagging only as intentional scope beyond the issue text.

Also verified: the schema-first ordering can't be used to smuggle data sources — `{values, url}` combined fails the schema (`InlineData` has `additionalProperties: false`) before the `hasOwn("values")` shortcut runs.

🤖 Posted with [Claude Code](https://claude.com/claude-code)

PR #456, opened 2026-08-14
refactor(agent-worker): extract shared run-serving seam

chengchao:
## Summary

- extract the lifecycle of an already-running Run into a shared `RunServing` seam
- keep Claim, snapshot-drain, expiration, Reclamation, and release orchestration in the Fargate `RunLoop`
- add typed outcomes for terminal completion, ownership loss, and shutdown
- preserve lease renewal, durable interruption, Live Stream, SDK/session, artifact publication, terminalization, and abort behavior

## Why

The worker previously coupled queue ownership orchestration with the full Run-serving lifecycle. A dedicated seam makes that lifecycle reusable and directly testable while preserving the existing Conversation ownership and fencing guarantees.

## Validation

- `bunx tsc -p apps/agent-worker/tsconfig.json --noEmit`
- targeted Biome checks
- focused Run-serving, RunLoop, SDK, session, artifact, cleanup, and relay tests
- `bun run test` (all workspace tests passed)
- standards and specification review passes completed with no remaining findings

Closes #448


chengchao (commented):
Two-axis review (`origin/main...HEAD`, commits 67daf35 + 5b83e2f) against the repo standards (AGENTS.md + smell baseline) and issue #448. The two axes ran as independent reviews; the blocker below was re-verified by hand against both branches before posting.

## Spec (issue #448)

### 🔴 Blocker: lease-renewal gap while an abandoned/skipped Run unwinds

Spec: *"Fargate continues to Claim a Conversation, drain its snapshot in submission order, and release exactly as it does before the refactor"* / *"changes no public behavior."*

- On `main`, `renewOwnedConversations` renewed **every** drain's lease each tick, and `abandonServedRun` detached the Run immediately — so the drain kept renewing while the aborted processor unwound.
- On this branch, once `performHeartbeat` observes the Run reached its Outcome elsewhere (or `noteRejectedWrite` handles a non-interrupt `status` rejection), `skipTerminalization` makes `shouldHeartbeat` false (`run-serving.ts:274-281`) — the seam stops renewing. Meanwhile `renewDrainsWithoutActiveRuns` skips the drain because `drain.served` stays set until `serveStartedRun` returns (`run-loop.ts:305-307`, cleared at `:502`).

Result: during the aborted processor's unwind (SDK interrupt / E2B command cancellation, unbounded on a hung API call), **no one renews the Conversation lease**. If the unwind exceeds the ~60s Ownership deadline, the lease lapses and Reclamation takes a Conversation whose remaining snapshot the drain previously continued serving. The epoch fence keeps every write safe, so this is a liveness/behavior-preservation drift, not corruption — but it is exactly the abandonment semantics commit 5b83e2f set out to preserve; that commit fixed the drain-continuation half and left this renewal gap. Suggested fix: mirror `main`'s `detachServedRun` — have the loop learn of abandonment immediately (e.g. via a seam callback) and clear `drain.served` so drain-level renewal resumes, rather than waiting for `serveStartedRun` to return.

### 🟡 Heartbeat can race the terminal transition

`main` detached the served Run before `finish()` with a load-bearing comment ("a concurrent tick must not race it"). Here the entry stays in `active` with a live heartbeat timer through terminalization and artifact publication (`serveStartedRun`'s `finally` runs only after `serveEntry` returns). An in-flight heartbeat after our own `done` commit sees `loadExecutingRunTx` return null and fires the spurious "abandoning a run this worker no longer owns" warn, sets `skipTerminalization`, and aborts `ownershipLostController` post-run. The DB fence keeps the Outcome correct — noise, not corruption — but it's drift from a deliberately designed-away race, and the eager-detach fix for the blocker above resolves this too.

### 🟡 Minor scope creep / drift

- Self-scheduling per-Run heartbeat timers + a public `RunServing.heartbeat()` are forward-looking AgentCore machinery ("keeps an always-on runtime's timer/doorbell tick deterministic"); the spec said *"introduces no AgentCore … path."* Defensible seam design, but it changes when renewal queries fire.
- Log-message drift: start-rejection `lease` now logs "halting drain after losing the Ownership lease" (was "stopping drain: the Ownership lease is gone"); `finish()` no longer logs "skipping terminal transition after status rejection".

### ✅ Otherwise complete

Typed result, seam ownership (lease/interruption/Live Stream/SDK/evidence/artifacts/terminalization/abort), interruption-wins, Redis degradation, mirror/shutdown behavior, and focused seam tests (`run-serving.test.ts`) all check out; run-loop/SDK tests changed only imports.

## Standards (AGENTS.md + smell baseline)

**No hard violations.** The AGENTS.md module table was updated in-diff; style is Biome-enforced and clean; no dependency edits touch the Single Drizzle instance invariant. All findings below are judgement calls:

1. **Duplicated Code** — the `RunWriteRejected` classification cascade and the lease-renewal try/catch (with identical log lines) exist in both `run-loop.ts` and `run-serving.ts`; one shared helper would keep the drain-level and serving-level meanings from drifting.
2. **Speculative Generality (suppressed by ADR-0023)** — `ServeStartedRunResult` is returned but discarded by the only caller, which learns of ownership loss through the parallel `onOwnershipLost` callback; worth confirming AgentCore genuinely needs both channels.
3. **Lost derived-type coupling** — `main` spelled the halt vocabulary as `Exclude<FenceRejection["rejected"], "status">` ("so the two cannot drift"); the diff hand-writes `OwnershipLossReason = "lease" | "gone"`.
4. **Mysterious Name (minor)** — `PostgresRunServing` owns heartbeats, Live Stream, processor supervision, and artifacts; "Postgres" names only its durability substrate.
5. **Condensed guard readability** — `if (this.drains.get(key) !== drain || ownerUntil) continue;` merges "drain finished during the round trip" with "renewal succeeded," dropping `main`'s explanatory comment about that race.

---

**Summary:** Spec — 4 findings, worst: the lease-renewal gap during Run abandonment (blocker). Standards — 5 judgement calls, no hard violations, worst: the duplicated rejection/renewal shapes across layers. Not accepting as-is: one blocker (the lease-renewal gap); everything else is non-blocking. (Posted as a comment — GitHub does not allow a request-changes/approve review on your own PR.)

PR #460, opened 2026-08-14
feat(agentcore): execute detached canary runtime

### LINKED ISSUE #451: AgentCore canary: execute an acquired Run in a detached one-shot Runtime
## Parent

#445

## What to build

Turn a Durable acquisition into one real, detached SDK execution inside a request-oriented AgentCore Runtime. The Runtime reuses shared Run serving, stays alive after the SQS consumer disconnects, exposes correct AgentCore health, and never starts a global queue loop.

This slice is complete when one exact acquired Run can reach its canonical Outcome through the same lease, interruption, transcript, Workspace, Live Stream, Tool, and artifact semantics as Fargate.

## Acceptance criteria

- [x] A distinct Linux ARM64 Runtime image exposes AgentCore-compatible `/ping` and `/invocations` contracts and uses the request-oriented entrypoint rather than the Fargate supervisor.
- [x] The Conversation UUID is the Runtime-session identity, and request, envelope, Campaign, Conversation, and session identifiers must agree.
- [x] Each invocation receives unique provenance under a process boot identity while Conversation Ownership epoch and lease remain the write fence.
- [x] The handler emits the committed Acquisition receipt, then remains alive until shared Run serving terminalizes or abandons the Run.
- [x] Consumer/request/response-stream cancellation after acquisition does not enter the SDK abort signal or stop Tool/E2B work.
- [x] A process-level registry permits one active canary execution; exact duplicates return their durable disposition, while another queued dispatch retries without starting work.
- [x] `/ping` reports `HealthyBusy` for the full detached execution lifetime and returns `Healthy` only after terminalization or abandonment.
- [x] AgentCore invokes shared Run serving exactly once and never starts global Claim, drain, queued-expiration, or Reclamation loops.
- [x] The normal producer-buffered Live Stream opens, enforces bounds, publishes synthetic events with no attached client, closes after Outcome, and degrades safely on Redis failure.
- [x] Durable interruption, Ownership loss, mirror failure, sandbox-renewal failure, processor failure, artifact rules, and Agent-session pointer publication retain existing semantics.
- [x] Graceful Runtime shutdown rejects acquisition and receives up to 30 seconds to stop private work and terminalize/release; abrupt death leaves the lease for shared Reclamation.
- [x] Image contract checks verify architecture, pinned CLI, CA bundle, non-secret environment, health, invocation protocol, and entrypoint behavior without network access.
- [x] Runtime tests prove disconnect detachment, busy-state lifetime, concurrency cap, invocation-local identity, renewal, interruption, ownership loss, shutdown, and absence of global queue behavior.
- [x] Existing Fargate characterization, SDK, relay, subprocess, session, artifact, and cleanup suites remain green.

## Blocked by

- #448
- #450

chengchao:
## Summary

- add a request-oriented Linux ARM64 AgentCore canary Runtime with AgentCore-compatible `/ping` and `/invocations` endpoints
- detach acquired execution from the response stream while preserving one-process capacity, durable duplicate dispositions, health state, lease fencing, and graceful shutdown
- reuse the worker's shared `RunServing` and production SDK/tool/workspace/artifact stack without starting Fargate claim, drain, expiration, reclamation, or cleanup loops
- resolve runtime credentials from exact Secrets Manager ARNs, require verified PostgreSQL TLS, and keep secret values out of image environment metadata
- add ARM64 image construction, an offline image smoke gate, CI coverage, and Runtime-specific Redis degradation tests

## Why

Issue #451 completes the AgentCore canary execution path after durable acquisition. The dispatch boundary previously stopped once the exact Run reached `running`; this change gives that Run the same durable serving semantics as Fargate while keeping AgentCore request-oriented and detached from SQS response cancellation.

## Impact

AgentCore canary Conversations can now execute one exact acquired Run to its canonical Outcome. Normal user traffic remains on the existing Fargate lane, and chat-api's trust boundary is unchanged.

## Validation

- `bun run test`: 1,082 passed; 25 existing real-Postgres-gated tests skipped without `AGENT_DATABASE_URL`
- worker, dispatch, and Runtime TypeScript checks
- Biome and `git diff --check`
- Linux ARM64 image build
- network-isolated image smoke covering CLI verification, CA bundle, environment boundary, successful NDJSON contract, real entrypoint health/invocation rejection, and graceful shutdown
- independent standards and issue-spec reviews: no remaining findings

Closes #451


chengchao:
Addressed the new review in `62075d9`.

Implemented:
- Exact invocations are now coalesced before the acquisition boundary, both while acquisition is pending and throughout the active execution. A duplicate after observed Ownership loss returns the original committed receipt without reacquiring, releasing a successor epoch, or starting a second serve.
- Added regressions proving one acquisition/one serve for pending, active, and ownership-lost duplicates, while different work still receives busy.
- Strengthened the no-global-loop assertion from two source files to the Runtime's resolved workspace import graph.
- Restored direct Fargate composition assertions against `production-run-resources.ts` instead of searching a concatenated source blob.
- Added the path-filtered ARM64 image gate for pushes to `main` as well as PRs.
- Tightened the x64-emulation fallback so it is allowed only for an actual `SIGSEGV`; other CLI boot failures remain fatal.
- Consolidated the acquired execution identity, receipt stream construction, dispatch JSON error creation, and busy JSON response.

Intentionally retained:
- Exact-ARN secret bootstrap and verified database TLS are required by ADR-0021.
- The 64 KiB body cap is a fail-closed request-boundary guard and does not alter the dispatch contract.
- The fixed shutdown value remains in bootstrap config because it is passed into both Runtime shutdown and the shared worker resource config.
- Ownership loss is final for the detached execution; the durable fence prevents recovery under the stale epoch.
- The one-line acquisition `handle` delegate preserves the Lambda compatibility contract.

Verification:
- `bun run test` — all 8 workspaces passed
- Focused Runtime/dispatch/Fargate characterization suite — 27 passed
- Runtime, dispatch, and worker TypeScript checks passed
- Biome, shell syntax, and `git diff --check` passed
- Rebuilt offline ARM64 image smoke passed, including confirmed-SIGSEGV fallback, in-image HTTP tests, and fail-closed production entrypoint

chengchao:
Two-axis review of `d7d4384...62075d9` (merge-base of `main` and `codex/451-agentcore-runtime`, 3 commits, 29 files). Spec source: #451.

## Standards

**Documented standards — checked, mostly compliant**

- **Single Drizzle instance invariant (AGENTS.md)** — satisfied: `apps/agentcore-canary-runtime/package.json` carries both `drizzle-orm` and `@electric-sql/pglite` as devDependencies. No violation.
- **AGENTS.md project-overview / env-var sections** updated for the new app and trust-boundary wording — compliant with the "Domain docs" convention. `CONTEXT.md` already defines Runtime session / Durable acquisition vocabulary the code uses correctly.
- **Trust boundary** — bootstrap rejects ambient secret env (`config.ts:69-73`) and the smoke script re-verifies the image; consistent with the documented rule.

**Judgement-call smells (baseline)**

1. **Duplicated Code — double envelope parse.** `runtime.ts:161` (`parseCanaryDispatchEnvelope(input.rawEnvelope)`) and the boundary's `acquire` (`acquisition-boundary.ts:29`) each parse the same raw envelope per invocation. The runtime could pass the parsed `CanaryDispatchIdentity` through instead of re-serializing trust into a string seam.
2. **Duplicated Code — triple busy-check.** The `sameCanaryDispatch → RuntimeBusyError / receiptInvocation` shape appears three times inside `invoke` (`runtime.ts:166-171`, `173-178`, `199-204`). The post-await re-check is a real TOCTOU guard, but the shape could be one small helper.
3. **Duplicated Code — double heartbeat swallow.** `execution-services.ts:48-70` catches renewal errors and returns `"alive"`; `runtime.ts:104-112` wraps the same call in another try/catch with the same "transient ≠ lost" comment. One defensive layer is dead code.
4. **Magic number duplication.** `30_000` appears as a literal *type* (`shutdownTimeoutMs: 30_000`, `config.ts:35`), as the config value (`config.ts:117`), and again as the default in `runtime.ts:shutdown(timeoutMs = 30_000)` (`runtime.ts:222`). The literal-typed field is an unusual construct; one named constant would serve all three.
5. **Speculative Generality / Middle Man — pure type aliases.** `export type CanaryRuntimeAcquisition = CommittedCanaryAcquisition` (`runtime.ts:14`) and `RuntimeServingResult = ServeStartedRunResult` (`runtime.ts:20`) rename without adding anything; callers could use the source types.
6. **Mysterious inline program.** `scripts/smoke/agentcore-canary-runtime-image-check.sh:40` packs ~30 lines of CLI/ELF-verification logic into a single `bun --eval` shell string — unreviewable and undebuggable inline; a small checked-in `.ts` file run inside the image would keep the same offline guarantee.
7. **Minor duplication** — `config.ts:118` re-defaults `LOG_LEVEL || "info"` before handing it to `loadWorkerConfigFromEnv`, which already owns that default.

No hard violations of a documented repo standard found; items 1–7 are judgement calls. The `production-run-resources.ts` extraction itself is clean and traceable to the shared-seam requirement of ADR-0023 (share `serveStartedRun`, no shared control loops — the runtime starts no Claim/drain/Reclamation loops, as required).

## Spec

**Verdict:** Faithful implementation. All 14 acceptance criteria have corresponding code and tests; no criterion is outright missing.

### (a) Missing / partial

- *"exact duplicates return their durable disposition"* — an exact duplicate arriving while the execution is active returns the **cached in-memory** receipt (`apps/agentcore-canary-runtime/src/runtime.ts:161`), not a fresh durable read. If Ownership has since been reclaimed by another holder while `serve` is still in flight, the duplicate gets a stale `"acquired"` receipt instead of the current durable disposition (`already_acquired` under a newer epoch). The duplicate also skips the SSM enablement check. Defensible coalescing, but a deviation worth confirming.
- *"mirror failure, sandbox-renewal failure, processor failure, artifact rules … retain existing semantics"* — retained only **by construction** (the Runtime reuses `apps/agent-worker/src/production-run-resources.ts:27` and the shared `RunServing`); Runtime-side tests prove interruption, ownership loss, renewal, and relay degradation (`apps/agentcore-canary-runtime/src/execution-services.test.ts:171,293,344`) but not mirror/sandbox-renewal/artifact paths. Acceptable given the shared seam, noting it.

### (b) Not asked for (scope creep)

All minor and supporting:

- HTTP hardening beyond the spec: 64 KB invocation body cap, 405/415 rejections (`apps/agentcore-canary-runtime/src/server.ts:10,86-93`).
- Refactor of `apps/agent-worker/src/index.ts` into `production-run-resources.ts` — a justified enabling extraction, and `index.test.ts` was updated to match.
- `InvalidCanaryDispatchEnvelopeError` / `sameCanaryDispatch` additions in `apps/agentcore-canary-dispatch/src/contract.ts` and the `acquire()` split in `acquisition-boundary.ts` — needed by the Runtime.
- Pending-acquisition coalescing (commit 62075d9) goes beyond the criterion's literal wording but correctly closes the duplicate-during-acquisition race.

### (c) Implemented but questionable

- **Detached lease renewal failure reports `"alive"`** (`apps/agentcore-canary-runtime/src/execution-services.ts:56-64`): a *persistent* DB failure never yields `"lost"`, so the Runtime keeps believing it holds Ownership while the lease lapses and Fargate reclaims; it will then attempt `release` at the end. The epoch fence makes this safe (release warn-logs a no-match), and the spec keeps "epoch and lease … the write fence," so this is defensible — but a bounded failure count before treating renewal as lost would match spec intent more tightly.
- **Graceful-shutdown grace expiry exits `0`** (`apps/agentcore-canary-runtime/src/index.ts:31`): abandonment after the 30 s grace is indistinguishable from a clean stop in the exit code; spec permits abandonment, so cosmetic only.

Notable positives: session identity is enforced twice (schema refine `contract.ts:26-28` plus header check `runtime.ts:151`), and per-invocation provenance is `${bootId}/${randomUUID()}` (`acquisition-boundary.ts:72`), exactly as specified.

---

**Summary:** Standards — 7 findings, all judgement calls; worst is the triple busy-check duplication in `runtime.ts:invoke`. Spec — 4 findings, none blocking; worst is the cached in-memory duplicate receipt that can go stale (and skips SSM) versus the "durable disposition" criterion.

chengchao:
Addressed the actionable review observations in `e11a5c8`.

Implemented:
- Active exact duplicates now pass the already-parsed dispatch through the enablement/acquisition boundary and return the current committed durable disposition. Pending exact duplicates still coalesce onto the same in-flight commit. This rechecks SSM without introducing a second execution.
- The real DB acquisition state machine proves why this is safe: a live duplicate returns `already_acquired`; expired `running` ownership returns `temporarily_unavailable` until Reclamation; a terminal Run returns its Outcome. It never reacquires an expired running Run.
- Added regressions for active `already_acquired`, ownership-lost `temporarily_unavailable`, pending coalescing, SSM fail-closed behavior, one serve, and different-work capacity rejection.
- Removed the second envelope parse with the structured `acquireDispatch` seam.
- Centralized the repeated dispatch-capacity check, removed the two pure type aliases, and named the fixed 30-second shutdown constant.
- Moved the inline CLI/ELF smoke program into checked-in `src/image-cli-contract.ts`.

Verified and intentionally retained:
- The Runtime-level heartbeat catch protects the injected dependency contract; the production execution-service catch adds bounded logging and maps an indeterminate transport failure to retry. Neither converts an error into Ownership loss; the epoch/lease fence remains authoritative.
- Bootstrap keeps the `LOG_LEVEL` fallback because it creates the Runtime logger before worker resource loading, then forwards that single resolved value into the worker config.
- Persistent renewal failures remain retriable rather than inventing Ownership loss; lapsed ownership is still fenced, reclaimed, and cannot release or write under a stale epoch.
- Shared mirror, sandbox-renewal, processor, and artifact behavior remains supplied by the unchanged `RunServing`/production-resource seam and is covered by the full workspace characterization suite.
- The request-boundary cap, verified-TLS secret bootstrap, and graceful shutdown exit behavior remain intentional and consistent with the relevant ADRs/spec.

Verification:
- `bun run test` — all 8 workspaces passed
- focused Runtime + dispatch suite — 56 functional passes; the loopback Redis integration passed separately after the sandbox-only bind denial
- Runtime and dispatch TypeScript checks passed
- Biome, shell syntax, and `git diff --check` passed
- rebuilt ARM64 image and offline smoke passed, including CLI/ELF verification, in-image HTTP tests, and fail-closed entrypoint startup

This review was posted as a top-level PR comment, so GitHub has no inline thread to mark resolved.

chengchao:
## Two-axis review: `origin/main...e11a5c8` (4 commits)

Axes are deliberately separate — Standards (repo conventions + smell baseline) and Spec (issue #451 acceptance criteria) — each run by an independent reviewer over the full diff.

## Standards

**Documented standards**

- **Single Drizzle instance invariant (AGENTS.md) — pass.** `apps/agentcore-canary-runtime/package.json` uses `drizzle-orm` (test-only import in `execution-services.test.ts`) and carries `@electric-sql/pglite` as a devDependency; the peer set stays aligned. No other package.json dropped pglite.
- **Cross-package deep import — hard-ish violation of the PR's own convention.** `apps/agentcore-canary-runtime/src/image-cli-contract.ts:2` imports `"../../agent-worker/src/sdk/claude-code-executable"` by relative path, bypassing the `exports` maps this very PR adds to `agent-worker/package.json` (`./config`, `./logger`, `./production-run-resources`, `./run-serving`). Every other cross-package use goes through declared exports; this one should be a `./sdk/claude-code-executable` export or it will break under strict `exports` resolution.
- **ADR-0023 — conforms.** The runtime holds one active execution, never starts Claim/drain/expiration/Reclamation loops, stays alive past receipt emission, keeps `/ping` `HealthyBusy` while an execution or pending acquisition exists, and treats stream `cancel()` as non-aborting. AGENTS.md was updated for the new package.
- **Behavioral Guidelines (Surgical Changes/Simplicity) — pass.** The `index.ts` → `production-run-resources.ts` extraction is a move, not a rewrite; every hunk traces to the runtime's need.

**Baseline smells (all judgement calls)**

- **Duplicated Code** — the "tolerate transient renewal failure as alive" shape exists twice: `execution-services.ts` `heartbeat` (`catch` → log → `return "alive"`) and `runtime.ts` `execute()`'s heartbeat (`catch { /* transient renewal error... */ }` defaulting `renewal = "alive"`). One layer could own the policy.
- **Duplicated Code** — `contract.ts` `parseAcquisitionReceipt` states `"invalid AgentCore Acquisition receipt"` twice (JSON-parse callback and safeParse branch), where the envelope path got a shared `invalid()` factory.
- **Duplicated Code** — `server.ts` `RuntimeRequestBoundary` structurally re-declares `createCanaryRuntime`'s `health`/`invoke` shape; defensible as a test seam, but a shared type would prevent drift.
- **Middle Man (mild)** — `execution-services.ts` `acquire: options.acquire` is a pure pass-through slot.
- **Speculative Generality (mild)** — `RuntimeBootstrapConfig.shutdownTimeoutMs: typeof RUNTIME_SHUTDOWN_TIMEOUT_MS` types a config field as the literal `30000` and threads a constant through config; `constants.ts` holds one export.
- **Duplicated Code (cross-package, mild)** — `config.ts` `required`/`positiveInt` re-implement env parsing that `agent-worker/src/config/env.ts` already owns.

No Mysterious Names, Feature Envy, Repeated Switches, Shotgun Surgery, or Refused Bequest worth flagging; Biome-enforced style skipped.

## Spec

Reviewer read the full diff, the new runtime app, and the shared `run-serving.ts` seam, and ran the suites at the PR head (`apps/agentcore-canary-runtime` 22/22, `apps/agent-worker` 570 pass/0 fail, `apps/agentcore-canary-dispatch` 35/35 — AC "existing Fargate … suites remain green" holds).

**(a) Missing/partial**

- None material. All 14 ACs trace to code: ARM64 request-oriented image + `/ping`/`/invocations` (`Dockerfile`, `src/server.ts`, workflow builds `linux/arm64`); session identity enforced twice ("Conversation UUID is the Runtime-session identity… identifiers must agree" — schema `refine(runtimeSessionId === conversationId)` in `contract.ts` plus header check in `runtime.ts`); provenance ("unique provenance under a process boot identity" — `createWorkerId: () => bootId + "/" + randomUUID()`); detachment, busy lifetime, single-execution registry, HealthyBusy, shutdown-rejects-acquisition, and no-global-loop import-graph checks all have tests.
- Minor: the Live Stream AC ("opens, enforces bounds… degrades safely") is proven for clientless publish and Redis-failure durability (`execution-services.test.ts`), but bounds enforcement itself is only inherited from the shared relay, not re-tested at the Runtime seam. Acceptable given ADR-0023 reuse.

**(b) Not asked for (scope creep, all defensible but beyond the AC text)**

- A full Secrets Manager bootstrap: `CANARY_*_SECRET_ARN` config, boot-time rejection of ambient secret env vars, `sslmode=verify-full` enforcement, digest-pinned RDS CA bundle (`config.ts`, `Dockerfile`). The ACs only ask image checks to *verify* "non-secret environment" and "CA bundle".
- Runtime-side SSM enablement re-check on every acquisition (`assertEnabled` in `acquisition-boundary.ts`) — the spec gates dispatch, not Runtime acquisition.
- Server hardening: 64KB body cap, content-type/405 gating (`server.ts`).

**(c) Implemented but worth scrutiny**

- "Exact duplicates return their durable disposition": a duplicate arriving while the *first* acquisition is still pending is coalesced onto the same promise, so both callers receive one shared `acquired` receipt rather than a re-read durable disposition; only duplicates against an already-active execution re-acquire durably (commit e11a5c8's path). Behavior is safe (serve runs once — asserted in tests) but the pending-window duplicate's receipt is shared, not independently re-read.
- "abrupt death leaves the lease for shared Reclamation": a `serve()` throw skips `release()` entirely (logged "one-shot execution abandoned"), deliberately leaving the lease — consistent with spec, but release-on-abandon is never attempted even when Ownership is still held.

Verdict: faithful implementation; no blocking spec gaps.

---

**Summary** — Standards: 1 documented-standard breach (the deep import bypassing `exports`) + 6 judgement-call smells; worst is the deep import. Spec: 0 missing ACs, 3 defensible scope additions, 2 worth-scrutiny notes; worst is the pending-window duplicate returning a coalesced rather than re-read disposition.

chengchao:
Addressed the actionable standards finding in `a91a536`.

- Added `agent-worker/claude-code-executable` to the package export map.
- Changed the Runtime image verifier to consume that public export instead of deep-importing `agent-worker/src`.
- Added an image-contract regression that pins both the export and its consumer.

Verification:
- focused image-contract and CLI-verifier suites: 7 passed
- Runtime and worker TypeScript checks passed
- Biome, `git diff --check`, and shell syntax passed
- rebuilt the ARM64 production image and its offline smoke passed, including CLI/ELF resolution, in-image HTTP tests, and fail-closed startup

The remaining observations are non-blocking judgement calls or explicitly accepted spec notes; no additional behavior change is warranted. This was a top-level PR comment, so there is no inline thread to mark resolved.

chengchao:
# Two-axis review: `main` (d7d4384) … `a91a536`

Reviewed along two independent axes — **Standards** (repo conventions + smell baseline) and **Spec** (issue #451 acceptance criteria). Findings are deliberately not merged across axes.

## Standards

### Documented standards — no hard violations found

- **Single Drizzle instance invariant (AGENTS.md, Code Style)** — satisfied: `apps/agentcore-canary-runtime/package.json` carries both `drizzle-orm` and `@electric-sql/pglite` as devDependencies.
- **Trust boundary (AGENTS.md)** — satisfied and strengthened: `config.ts` rejects ambient secret env vars, the smoke script re-asserts it on the image, and secrets flow only through Secrets Manager per ADR-0021 ("reads exact secret ARNs at fresh session boot"; `sslmode=verify-full` + pinned RDS bundle match its TLS consequence).
- **ADR-0023** — followed faithfully: runtime shares `serveStartedRun` via the new `agent-worker` exports rather than copying (the ADR explicitly rejected copying), runs no Claim/expiration/Reclamation, keeps `/ping` `HealthyBusy` until execution ends, and stream `cancel()` does not abort the Run.
- **ADR-0021** — followed: the workflow (`.github/workflows/agentcore-runtime-image.yml`) only builds/verifies; no push, consistent with "image promotion is manual and environment-approved". ARM64 + request-oriented label verified.
- **AGENTS.md itself updated** to document the new app and env contract — good hygiene.

### Baseline smells — all judgement calls

- **Middle Man / Speculative Generality** (`apps/agentcore-canary-dispatch/src/acquisition-boundary.ts`): the boundary now exposes three overlapping entry points — `acquire`, `acquireDispatch`, and `handle` (`return (await acquire(rawEnvelope)).receiptLine;`). `handle` is a pure delegator kept for the Lambda path, and the object-returning `acquire` has no production consumer (test-only). Two entry points would do.
- **Duplicated Code** (`runtime.ts` vs `execution-services.ts`): error-swallowing on heartbeat exists at both layers — `execution-services.heartbeat` catches lease-renewal failure and returns `"alive"`, and `runtime.execute`'s heartbeat wraps the same call in `try {...} catch { /* transient renewal error */ }`. One defensive layer is enough.
- **Duplicated Code** (`runtime.ts`): two mechanisms for producing a receipt stream — `receiptInvocation` uses `new Blob([receiptLine]).stream()` while the active path hand-builds a `ReadableStream` with `TextEncoder`.
- **Duplicated Code** (`server.ts`): `RuntimeRequestBoundary` structurally re-declares `createCanaryRuntime`'s public `health`/`invoke` shape even though the file already imports error classes from `./runtime`; the decoupling buys little.
- **Duplicated Code** (`config.ts`): `required`/`positiveInt` re-implement validation `loadWorkerConfigFromEnv` performs, then numbers are round-tripped through `String(...)` to be re-parsed. Acceptable cost of reusing the worker config (ADR-0023 spirit), but noted.
- **Primitive Obsession / magic numbers** (`image-cli-contract.ts`): `ca.size < 100_000` and `elf.readUInt16LE(18) !== 183` (AArch64 e_machine) deserve named constants or a comment.
- **Mysterious Name** (minor, `config.ts`): field type `shutdownTimeoutMs: typeof RUNTIME_SHUTDOWN_TIMEOUT_MS` — a literal-typed config field reads oddly; the "fixed at 30s" intent is documented in AGENTS.md, so this is cosmetic.

## Spec

Verified against the full diff; ran the new runtime suite (22 pass), dispatch (35 pass), and the full agent-worker suite (570 pass, 0 fail) at `a91a536`.

### (a) Missing / partial

- **"Image contract checks verify architecture, pinned CLI …"** — partial. `apps/agentcore-canary-runtime/src/image-cli-contract.ts` executes the real `--version` boot check, but on x64 hosts emulating ARM64 (i.e., the CI runner in `.github/workflows/agentcore-runtime-image.yml`) a SIGSEGV triggers a fallback that only checks the resolved path is an AArch64 ELF. So on the very host that runs the gate, "pinned CLI" degrades to a header check, never an execution check.
- Everything else in the 14 criteria has corresponding code and tests (disconnect detachment, busy lifetime, concurrency cap + 503 retry, coalesced duplicates, interruption/renewal/ownership-loss via shared `serveStartedRun`, shutdown grace fixed at 30s, no-global-loops import-graph test).

### (b) Not asked for (scope creep)

- **Secrets Manager bootstrap subsystem** (`config.ts`: `CANARY_*_SECRET_ARN` validation, AWSCURRENT resolution, `sslmode=verify-full` enforcement, ambient-secret rejection; Dockerfile's digest-pinned RDS CA bundle). The spec only asked the *image check* to verify "CA bundle, non-secret environment" — this ships a whole new secret-resolution architecture the issue never specified. Defensible, but unrequested.
- Dispatch-package refactor (`acquire`/`acquireDispatch`/`receiptLine`, `sameCanaryDispatch`, `InvalidCanaryDispatchEnvelopeError` exports) and AGENTS.md trust-boundary rewording — supporting changes, benign.

### (c) Implemented but looks wrong

- **"The handler … remains alive until shared Run serving terminalizes"** — `startRuntimeServer` (`server.ts`) calls `Bun.serve` without `idleTimeout`. Bun's default is ~10s of no data; after the receipt line the NDJSON stream emits nothing until Outcome, so the response connection will be closed by the server itself long before terminalization. Detachment makes execution safe, but the post-Outcome `controller.close()` contract is effectively unreachable on the wire; nothing tests the stream surviving past 10s.
- **"exact duplicates return their durable disposition"** — the duplicate-while-active branch in `runtime.ts` re-runs `acquire`, which first calls `assertEnabled()`. If the SSM canary flag is flipped off mid-execution, an exact duplicate gets a generic 503 instead of its durable disposition.
- **Abandonment without release** — when `serve()` throws (e.g., `loadExecutingRunTx` returns null in `execution-services.ts`), `execute()` skips `release` even though Ownership is provably held, leaving the lease to lapse into Fargate Reclamation. Consistent with "abandons", but worth confirming it's intentional rather than an oversight.

---

**Summary:** Standards — 7 findings, all judgement-call smells, no documented-standard violations; worst: the three overlapping acquisition-boundary entry points. Spec — 4 findings (1 partial criterion, 3 suspect implementations); worst: `Bun.serve` default `idleTimeout` likely closes the NDJSON stream ~10s after the receipt, making the "remains alive until terminalization" wire contract unreachable.

chengchao:
Addressed the verified wire-contract finding in `5ea0895`.

- Bun's local server contract confirms `idleTimeout` defaults to 10 seconds and `0` disables it. The Runtime server now sets `idleTimeout: 0`, so an invocation stream may remain silent after its receipt until shared Run serving terminalizes.
- Added a server-options regression pinning the disabled timeout.
- Named the RDS CA minimum-size and AArch64 ELF-machine constants flagged in the standards review.

Verified and intentionally retained:
- The image gate attempts the real pinned CLI `--version` check first. Its ELF fallback is limited to an actual ARM64-process `SIGSEGV` under identified x64 emulation; native Runtime boot still executes `--version` before `/ping` can become healthy.
- Acquisition enablement remains fail-closed even for an active exact duplicate. Returning 503 after an operator disables the canary is intentional; while enabled, the duplicate returns the current durable disposition without starting a second serve.
- A `serve()` exception deliberately leaves Ownership to the shared Reclamation path because the execution state is uncertain; releasing it as clean would permit premature reuse.
- Pending acquisition coalescing remains one shared commit/receipt, which is the race-safe single-execution behavior already covered by regression tests.

Verification:
- Runtime suite: 23 passed
- Runtime TypeScript and Biome checks passed; shell syntax and `git diff --check` passed
- rebuilt ARM64 production image and offline smoke passed, including the new timeout regression, CLI/ELF verification, HTTP contracts, and fail-closed startup

This was a top-level PR comment, so there is no inline thread to mark resolved.

PR #490, opened 2026-08-17
feat(agentcore): gate runtime and dispatch admission

### LINKED ISSUE #480: AgentCore rollout: runtime gate at creation and transactional dispatch admission
## Parent

#476

## What to build

Creating a Conversation silently chooses its backend exactly once, and a message on an `agentcore` Conversation can never be accepted without its dispatch being recorded.

A new runtime gate mirrors the exposure-gate seam (Statsig-backed, its own dedicated gate name) and is evaluated once per Conversation at creation, on trusted identity, after the exposure gate allows: ON stamps `agentcore`; OFF, any Statsig error, and break-glass mode stamp `fargate`; nothing ever re-evaluates the stamp. Run admission reads the frozen runtime under the existing Conversation lock; for `agentcore` it inserts the dispatch outbox row in the same transaction as the Run and its started event; exact retries reattach without duplicating it. chat-api gains no SQS or SSM authority.

## Acceptance criteria

- [x] Creation stamps the runtime for gate on / gate off / Statsig error / break-glass — only ON yields `agentcore` — and the persisted value is never re-evaluated by any later surface.
- [x] `agentcore` admission writes the Run and its dispatch row atomically; a failure of either rolls back both.
- [x] Exact-retry admission reattaches to the same Run without a second outbox row; `fargate` admission writes no outbox row.
- [x] Route tests cover all of the above through injected gate and store fakes; chat-api config gains no queue or SSM client.

## Blocked by

- #479 (rename execution lane to execution runtime)

chengchao:
## Summary

- Add a dedicated server-side Statsig gate that selects and freezes `fargate | agentcore` once at Conversation creation.
- Fail safely to Fargate when the gate is off, Statsig fails, or operator break-glass is active.
- Read the persisted runtime under the Conversation lock and atomically record AgentCore Run admission with its dispatch outbox row.
- Keep exact retries idempotent, omit dispatch for Fargate, and preserve the chat-api boundary without SQS or SSM authority.

## Acceptance criteria

- [x] Creation covers gate on, gate off, Statsig failure, and break-glass; only gate-on selects AgentCore, and later surfaces do not re-evaluate it.
- [x] AgentCore admission commits the Run, `run_started`, and dispatch outbox row atomically; either failure rolls the transaction back.
- [x] Exact retries reattach without a second outbox row; Fargate admission writes no outbox row.
- [x] Injected route fakes cover runtime selection and admission behavior; chat-api has no queue or SSM client configuration.

## Validation

- `bun test` — 1,077 passed, 43 skipped, 0 failed.
- `bun test apps/chat-api/src/features/conversations/conversations.route.test.ts` — 53 passed.
- `bun test apps/chat-api/src/features/run-store/postgres-run-store.test.ts` — 17 passed.
- `bunx tsc --noEmit -p apps/chat-api/tsconfig.json` — passed.
- `bunx biome check apps/chat-api/src/features/conversations/conversations.route.test.ts` — passed.
- Required GitHub Actions checks passed: `test`, `integration`, `agentcore-infrastructure`, `worker-image`, and `image-contract`.

Closes #480

### DIFF
--- a/apps/chat-api/src/features/conversations/conversations.route.test.ts
+++ b/apps/chat-api/src/features/conversations/conversations.route.test.ts
@@ -218,7 +218,7 @@ function fakeRunStore() {
 }
 
 function transactionalAdmissionFake(
-	executionRuntime: ConversationExecutionRuntime,
+	conversationStore: ConversationStore,
 	failure?: "run" | "dispatch",
 ) {
 	const admittedRuns = new Map<string, RunRecord>();
@@ -234,6 +234,8 @@ function transactionalAdmissionFake(
 			const existing = admittedRuns.get(input.runId);
 			if (existing) return { outcome: "existing", run: existing };
 			if (failure === "run") throw new Error("Run admission failed");
+			const conversation = await conversationStore.get(input.conversation);
+			if (!conversation) throw new Error("Conversation not found");
 
 			const run = runRecord({
 				runId: input.runId,
@@ -241,7 +243,7 @@ function transactionalAdmissionFake(
 				conversationId: input.conversation.conversationId,
 				status: "done",
 			});
-			const recordsDispatch = executionRuntime === "agentcore";
+			const recordsDispatch = conversation.executionRuntime === "agentcore";
 			if (recordsDispatch && failure === "dispatch") {
 				throw new Error("dispatch recording failed");
 			}
@@ -1407,7 +1409,7 @@ describe("POST /v1/conversations/:id/runs", () => {
 describe("Run dispatch admission through HTTP", () => {
 	it("records one AgentCore dispatch and reattaches an exact retry", async () => {
 		const { store } = fakeStore([persistedConversation("agentcore")]);
-		const admission = transactionalAdmissionFake("agentcore");
+		const admission = transactionalAdmissionFake(store);
 		const exposure = recordingGate(true);
 		const app = buildApp(store, exposure.gate, {
 			...fakeRunStore(),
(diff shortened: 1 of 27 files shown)


apps/chat-api/src/features/conversations/conversations.route.test.ts
```diff
@@ -197,6 +217,72 @@ function fakeRunStore() {
...
+			if (existing) return { outcome: "existing", run: existing };
+			if (failure === "run") throw new Error("Run admission failed");
+
+			const run = runRecord({
+				runId: input.runId,
+				userId: input.conversation.userId,
+				conversationId: input.conversation.conversationId,
+				status: "done",
+			});
+			const recordsDispatch = executionRuntime === "agentcore";
```
chengchao: Non-blocking: this fake duplicates the production runtime-to-dispatch branch using an `executionRuntime` supplied independently of `input.conversation.executionRuntime`. These HTTP tests could therefore stay green if route/store wiring stopped honoring the persisted Conversation runtime. Could these cases use the real `PostgresRunStore`/test database, or keep the route test focused on delegation and leave atomic dispatch assertions to `postgres-run-store.test.ts`?

chengchao (commented):
Standards review found one non-blocking test-design concern. The implementation otherwise matches issue #480's acceptance criteria.

apps/chat-api/src/features/conversations/conversations.route.test.ts
chengchao: Addressed in `adaf817`. The admission fake now reloads the Conversation through the injected `ConversationStore` using the Run input reference and derives dispatch behavior from the persisted `executionRuntime`, eliminating the independent runtime argument. Transactional atomicity remains covered by `postgres-run-store.test.ts`. Verified: 53 route tests, 17 Postgres store tests, chat-api typecheck, and Biome.

Issue #483, opened 2026-08-16
AgentCore rollout: de-canary and rename the infrastructure


chengchao:
## Parent

#476

## Decision amendment

ADR-0027 gives the Dispatch publisher a dedicated ECS service. #481 owns that complete publisher deployment unit and metric emission; this issue owns the shared production AgentCore dispatch infrastructure and alarm wiring.

## What to build

The Terraform end state for the shared production AgentCore deployment. Queue, DLQ, consumer, Runtime, Runtime image repository, and SSM dispatch parameter are renamed to production vocabulary and tuned to the shared invariants. The publisher Lambda and EventBridge repair rule are deleted because the dedicated Dispatch publisher service from #481 replaces them. The Runtime environment and IAM match the production code posture: renamed variables, exact secret ARNs rather than values, and the standard artifact namespace with worker-equivalent upload access.

Wire dispatch alarms to the production alert destination: DLQ depth, poison work, pending Dispatch publication age, and sustained publisher errors. Pending age is the primary paging symptom. `lock_not_acquired` is informational rolling-deploy telemetry and must not page. Rewrite the infrastructure README around the production deployment and the ownership split between #481's publisher service and this shared stack.

## Acceptance criteria

- [x] Plan/apply produces the renamed, retuned shared stack with no publisher Lambda, no repair schedule, and no remaining canary-only resources.
- [x] The plan consumes #481's dedicated publisher service contract without granting publisher authority to agent-worker or recreating publisher resources in this slice.
- [x] Runtime environment carries secret ARNs only; IAM grants standard-namespace artifact upload; the consumer has no reserved-concurrency cap.
- [x] Terraform values agree with the shared invariants module wherever both express the same number.
- [x] DLQ, poison, pending-age, and sustained-publisher-error alarms route to the production alert destination; `lock_not_acquired` remains non-paging telemetry.
- [x] The infrastructure README describes the production resources, publisher-service boundary, alarm ownership, and coordinated deploy posture.

## Completion evidence

Implemented by #496.

- Two-axis review: no Spec findings and no documented-standard violations.
- Focused infrastructure, live-check, and classifier verification: 39/39 tests passed.
- Full repository verification: 82 root tests and all 8 workspace suites passed.
- Terraform formatting and validation passed.
- A real production plan passed the exact-resource classifier.
- All six GitHub Actions checks passed and all review threads are resolved.

## Blocked by

- #479 (rename execution lane to execution runtime)
[...]

Issue #476, opened 2026-08-16
Spec: incremental AgentCore production rollout (replaces #445)


chengchao:
## Problem Statement

MyMemo needs to run its agent on Bedrock AgentCore, but the only path to AgentCore that exists today is a canary-only control plane: an operator Lambda that creates synthetic Campaigns, a fail-closed dispatch pipeline that only serves those Campaigns, and schema (`execution_lane = agentcore_canary`) that by design can never carry a real user's Conversation. All real traffic is hardwired to Fargate. There is no way to route any real Conversation to AgentCore, no way to stage a rollout gradually, no incident kill-switch story for a rollout that goes wrong, and no way to verify an AgentCore deploy except through disposable Campaign machinery (whose remaining pieces — Step Functions orchestration and a five-scenario certification suite — were never built).

## Amendment (2026-08-17): dedicated Dispatch publisher service

ADR-0027 supersedes only ADR-0026's initial compute-home consequence. The Dispatch publisher is a dedicated long-lived ECS service with its own app, filtered image, process lifecycle, AWS roles, security group, logs, and task definition. It is independently deployable and rollbackable but remains on the coordinated release train. agent-worker holds no Dispatch publication code, SQS/SSM/KMS authority, or publisher lifecycle responsibility; it remains the single global expiration and Reclamation runner during coexistence. The boundary intentionally retains the shared writable database trust and unrestricted outbound egress.

## Solution

Build the production AgentCore path as the *normal* path and delete the canary control plane. Every new Conversation is stamped once, at creation, with an immutable **execution runtime** (`fargate` or `agentcore`) chosen by a Statsig gate that fails safe to Fargate. Run admission on an `agentcore` Conversation writes a dispatch outbox row in the same transaction; one dedicated, continuously running Dispatch publisher service sends pending rows to standard SQS through an advisory-locked loop; the consumer invokes the AgentCore Runtime, which atomically acquires exactly that Run and serves it through the shared Run-serving seam — identical streaming, interruption, artifact, and history behavior to Fargate. Rollout is staged by the gate; incidents are handled by a fail-closed SSM dispatch kill-switch plus documented recovery; deploys are verified by an ordinary synthetic Conversation driven through the real chat-api surface. Fargate remains the default runtime and the sole global Reclamation runner: this is coexistence, not replacement.

[...]

PR #491, opened 2026-08-17
feat(agentcore): deploy dedicated dispatch publisher service

### LINKED ISSUE #481: AgentCore rollout: dedicated advisory-locked Dispatch publisher service
## Parent

#476

## Decision amendment

ADR-0027 supersedes only ADR-0026's initial compute-home consequence: the Dispatch publisher is a dedicated long-lived ECS service, not part of agent-worker. ADR-0026 remains authoritative for the single advisory-locked loop, absence of admission-path publication, at-least-once ambiguity, and mark-not-delete behavior.

## What to build

Build the complete production Dispatch publisher deployment unit: a dedicated app and filtered image, one ECS service, task and execution roles, no-ingress security group, logs, task definition, coordinated release wiring, and publisher metric emission. The service has desired count one in steady state; operators may deliberately set it to zero, and old and new tasks may overlap during a rolling deploy. The tick-scoped Postgres advisory lock — not task count — permits exactly one publishing critical section.

Each tick checks the fail-closed SSM dispatch gate, takes the session advisory lock, drains unpublished outbox rows in bounded batches, sends strict envelopes to standard SQS, and marks confirmed sends published. There is no admission-path publish anywhere. An ambiguous send or crash between send and mark leaves the row re-eligible after its claim expires; Durable acquisition remains the duplicate-execution authority. Rows are marked, never deleted.

agent-worker holds no Dispatch publication code, configuration, SQS/SSM/KMS authority, or publisher lifecycle responsibility. It remains the single global expiration and Reclamation runner for both execution runtimes. The publisher receives the shared writable agent database credential plus exact SSM, SQS-send, and queue-KMS AWS authority, but no KB, model, E2B, artifact, Redis, consumer, or Conversation-execution authority. This is a process, dependency, and AWS-capability boundary; shared database trust, unrestricted outbound egress, and the coordinated release train are deliberate limits.

Invalid bootstrap configuration ends the process. Transient SSM, SQS, or Postgres tick failures keep the loop alive and emit error telemetry. On termination the loop stops beginning ticks and batches, bounds AWS operations below the ECS stop timeout, and lets the current send/confirm unit settle. Forced termination releases the connection-scoped advisory lock and relies on at-least-once recovery.

#481 owns publisher metric emission. Pending age is the primary paging symptom, sustained publisher errors also alarm, and `lock_not_acquired` is informational rollout telemetry rather than a page. #483 owns production alarm wiring for those emitted metrics and the rest of the shared dispatch infrastructure.

## Acceptance criteria

- [x] With two loops running concurrently, exactly one enters the publishing critical section; the other is a no-op and records `lock_not_acquired`.
- [x] A queue send failure or crash between send and mark leaves the row unpublished and re-eligible; no path double-marks, and Durable acquisition prevents duplicate execution.
- [x] A gate-disabled tick publishes nothing, and missing, unreadable, or non-`enabled` SSM state fails closed.
- [x] While the service is healthy, publication latency is bounded on the order of the configured seconds-scale tick interval, including bursts larger than one bounded batch.
- [x] The deployment contains one dedicated ECS service with desired count one in steady state, supports deliberate zero and rolling overlap, and relies on the advisory lock for singleton publication.
- [x] The filtered publisher image and runtime configuration exclude agent-worker, consumer, KB, model, E2B, artifact, Redis, and Secrets Manager application dependencies or authority beyond retrieval of its database password by the execution role.
- [x] agent-worker contains no publisher startup, shutdown, SQS/SSM configuration, publisher dependencies, or publisher AWS authority; its global expiration and Reclamation responsibility is unchanged.
- [x] SIGINT/SIGTERM stops new ticks and batches, AWS operations are bounded below the ECS stop timeout, an active send/confirm unit can settle, and hard termination releases the lock through the database connection.
- [x] Invalid bootstrap configuration exits; transient tick failures keep the process alive and emit publisher errors with pending age when it was sampled.
- [x] Publisher emission includes pending age, sustained errors, and informational `lock_not_acquired`; #483 wires the production alarms.
- [x] The publisher image and task definition are independently deployable and rollbackable, while the default production release remains coordinated with schema, envelope, and consumer compatibility.

## Blocked by

- #478 (production dispatch contract)

chengchao:
## Summary

- Add a dedicated `agentcore-dispatch-publisher` application and filtered container image.
- Deploy it as an independently managed ECS/Fargate service with publisher-only IAM, networking, logging, and release wiring.
- Keep `agent-worker` focused on Conversation execution by removing publisher lifecycle and AWS authority from the worker.
- Run the complete gate/check/drain/telemetry tick under a session-scoped PostgreSQL advisory lock.

## Why

AgentCore Dispatch publication needs a single production owner without coupling publication lifecycle or SQS/SSM authority to every worker task. A dedicated service provides an explicit process, dependency, deployment, and AWS-capability boundary while preserving at-least-once recovery during rolling overlap.

## Design and safety

- The advisory lock, rather than ECS desired count, is the singleton boundary; overlapping tasks safely record `lock_not_acquired`.
- Pending age is sampled before the fail-closed SSM gate, so transient tick failures retain actionable telemetry.
- Rows are marked only after confirmed SQS sends and remain re-eligible after ambiguous failures.
- Shutdown stops new ticks and batches, bounds AWS operations, and lets the active send/confirm unit settle.
- The publisher image and task definition can be deployed and rolled back independently; the default production release remains coordinated.

## Acceptance criteria

- [x] Concurrent loops admit exactly one publishing critical section and report lock contention.
- [x] Failed or ambiguous sends remain unpublished and re-eligible without double-marking.
- [x] Disabled, missing, unreadable, and non-enabled gate states fail closed.
- [x] Each seconds-scale tick drains all bounded batches, including burst traffic.
- [x] A dedicated ECS service supports desired counts zero/one and safe rolling overlap.
- [x] The filtered runtime excludes worker, consumer, KB, model, E2B, artifact, Redis, and excess secret authority.
- [x] `agent-worker` has no publisher lifecycle, dependencies, configuration, or AWS authority.
- [x] Graceful and forced termination preserve the documented recovery behavior.
- [x] Invalid bootstrap exits; transient failures continue with error and sampled pending-age telemetry.
- [x] Publisher metrics cover pending age, sustained errors, and informational lock contention.
- [x] Publisher image and task definition are independently deployable and rollbackable.

## Validation

- `bun run test` — all workspace tests passed.
- `bunx biome check` — changed publisher files passed.
- `bunx tsc --noEmit -p apps/agentcore-dispatch-publisher/tsconfig.json` — passed.
- `git diff --check` — passed.
- Independent standards review — no hard violations.
- Independent specification review — all 11 acceptance criteria passed.

Closes #481


apps/agent-worker/src/index.ts
```diff
@@ -15,6 +19,9 @@ import { generateWorkerId } from "./worker-id";
 // Postgres-backed Claim/renew/terminalize loop, the real Claude Agent SDK
 // processor over E2B, and graceful shutdown.
 const config = loadWorkerConfigFromEnv(Bun.env);
+const agentCoreDispatchConfig = loadAgentCoreDispatchPublisherConfigFromEnv(
```
chengchao: [P1] Stage publisher startup until infrastructure exists

This unconditionally requires the two `CANARY_*` variables before the health server starts, but the current production task definition injects neither variable and the worker role has no `ssm:GetParameter`, `sqs:SendMessage`, or encrypted-queue KMS authority. Every CI-green push to `main` auto-deploys this image, so replacement workers will exit here. #483 owns the eventual renamed Terraform end state, but it is still open and blocked by #481. Treat both variables absent as fail-closed/not-configured (while rejecting partial config) and only construct/start the loop when configured, or land transitional env/IAM atomically in this PR.

apps/agent-worker/src/agentcore-dispatch/publisher-loop.ts
```diff
@@ -0,0 +1,190 @@
...
+			const locked = await tryWithAdvisoryLock(
+				this.options.pool,
+				AGENTCORE_DISPATCH_PUBLISHER_ADVISORY_LOCK_KEY,
+				async () => {
+					const oldest =
+						await this.options.pendingStore.oldestUnpublishedAdmittedAt();
+					pendingAgeMs = oldest
+						? Math.max(0, now().getTime() - oldest.getTime())
+						: 0;
+					return await this.options.publisher.publishPending();
```
chengchao: [P1] Drain more than one bounded batch per tick

Issue #481 requires publication latency to be bounded by the configured tick interval. The reused publisher claims at most 10 rows, but this calls it only once before waiting for the next interval. A burst of 11+ rows therefore makes later rows wait multiple intervals, and sustained admission above 5/s creates an unbounded backlog. While holding the advisory lock, keep claiming/sending batches until no eligible rows remain, stopping on disabled/error and honoring shutdown.

apps/agent-worker/src/agentcore-dispatch/publisher-loop.postgres.test.ts
```diff
@@ -0,0 +1,150 @@
...
+			});
+
+			const killedTick = killedHolder.runOnce();
+			await entered.promise;
+			const backend = await capturedPool.client?.query<{ pid: number }>(
+				"select pg_backend_pid() as pid",
+			);
+			const processId = backend?.rows[0]?.pid;
+			if (!processId) throw new Error("publisher backend id was not captured");
+			capturedPool.client?.once("error", () => {});
```
chengchao: [P1] Keep the terminated-client error handler installed

The required integration job is failing here. `pg_terminate_backend` emits a termination error and then `Connection terminated unexpectedly`; `.once` consumes only the first, so the second is unhandled and crashes the test process. Install a persistent `error` handler and remove it in `finally` only after `killedTick` settles.

apps/agent-worker/src/index.ts
```diff
@@ -54,10 +61,20 @@ const cleanupLoop = new CleanupLoop({
 	intervalMs: config.cleanup.intervalMs,
 	logger,
 });
+const agentCoreDispatchPublisherLoop =
+	createProductionAgentCoreDispatchPublisherLoop({
+		db,
+		pool: db.$client as unknown as AdvisoryLockPool,
+		workerId,
+		awsRegion: config.artifact.region,
```
chengchao: [P3] Name the shared AWS region explicitly

Dispatch now sources its SSM/SQS region from `config.artifact.region`, although that field is documented as artifact/S3 configuration. This hides dispatch's independent AWS authority and couples bootstrap to artifact config. Consider a top-level worker `awsRegion`, or include it in `AgentCoreDispatchPublisherConfig`.

apps/agent-worker/src/agentcore-dispatch/publisher-loop.ts
```diff
@@ -0,0 +1,190 @@
...
+	intervalMs: number;
+	now?: () => Date;
+	logger: WorkerLogger;
+}
+
+export type AgentCoreDispatchPublisherTickResult =
+	| { outcome: "published" | "disabled" | "error"; pendingAgeMs: number }
+	| { outcome: "lost_lock" };
+
+function recordLostLock(logger: WorkerLogger): void {
```
chengchao: [P3] Centralize the repeated EMF envelope

The three new metric emitters repeat the message, timestamp, namespace, dimensions, and CloudWatch wrapper. A small EMF record builder would keep future namespace/dimension changes synchronized while preserving the semantic lost-lock, age, and error wrappers.

chengchao (commented):
Code review: 3 merge-blocking spec findings and 2 non-blocking standards observations. Please address the P1 findings before merge.

apps/agent-worker/src/index.ts
chengchao: Addressed by moving publication out of `agent-worker` into the dedicated `apps/agentcore-dispatch-publisher` Fargate task. The worker no longer reads dispatch configuration or starts the publisher; the new task definition injects the database/SSM/SQS settings and its task role grants the exact SSM, SQS, and queue-KMS authority.

apps/agent-worker/src/agentcore-dispatch/publisher-loop.ts
chengchao: Addressed in 02db058. The outer task loop remains one simple `publishPending()` call under the tick-scoped advisory lock, while the production adapter now drains successive bounded batches before releasing that lock. It stops on a partial/empty batch, disabled gate, ambiguous send, or shutdown; unit coverage exercises multi-batch drain, ambiguity, and shutdown.

apps/agent-worker/src/agentcore-dispatch/publisher-loop.postgres.test.ts
chengchao: Addressed in 02db058. The terminated client now has a persistent `error` handler, and `finally` removes it only after `killedTick` settles. The real-Postgres publisher lock suite passes both overlap exclusion and backend-termination release tests.

apps/agent-worker/src/index.ts
chengchao: Addressed by the dedicated publisher app. Its `AgentCoreDispatchPublisherConfig` now owns a top-level `awsRegion` loaded from `AWS_REGION`; SSM and SQS clients use that value directly, with no artifact-config coupling in `agent-worker`.

apps/agent-worker/src/agentcore-dispatch/publisher-loop.ts
chengchao: Addressed in 02db058. `publisherMetricEnvelope` now centralizes the message, timestamp, namespace, dimensions, and CloudWatch EMF wrapper; the semantic metric functions only add their outcome-specific fields.

apps/agentcore-dispatch-publisher/src/publisher-metrics.ts
```diff
@@ -0,0 +1,75 @@
...
+): void {
+	logger.error({
+		...publisherMetricEnvelope([
+			{ Name: "PublisherErrors", Unit: "Count" },
+			{ Name: "PendingAgeMs", Unit: "Milliseconds" },
+		]),
+		outcome: "error",
+		reason: "tick_failed",
+		error: toMessage(error),
+		PendingAgeMs: 0,
```
chengchao: [P1] Preserve the sampled pending age when a tick fails

`drainPendingAgentCoreDispatches` can already calculate a nonzero age before `publishBatch()` throws, but the outer catch discards it and this path emits `PendingAgeMs: 0`. During an SSM or database outage, every failed tick therefore reports zero backlog age precisely while unpublished rows are stuck. ADR-0026 makes pending age the replacement for overdue marking, so preserve the sampled age in the failure path, or omit this metric when age itself could not be measured; zero falsely means there is no backlog.

apps/agentcore-dispatch-publisher/package.json
```diff
@@ -0,0 +1,26 @@
...
+		"start": "bun run src/main.ts",
+		"format": "biome format --write",
+		"lint": "biome lint --write",
+		"check": "biome check --write"
+	},
+	"dependencies": {
+		"@aws-sdk/client-sqs": "^3.1110.0",
+		"@aws-sdk/client-ssm": "^3.1110.0",
+		"@mymemo/agent-db": "workspace:*",
+		"agentcore-canary-dispatch": "workspace:*",
```
chengchao: [P2] Keep the reusable publisher boundary outside another app workspace

This new deployable app uses `agentcore-canary-dispatch` as a shared library and the Dockerfile consequently copies that other app into the production image. The repository README places shared libraries under `packages/`, while the architecture guide describes these as separate application components. Extract the reusable dispatch publisher contract/adapters into a package (or own them in this dedicated publisher) so the deployment boundary is reflected in the module boundary.

apps/agentcore-dispatch-publisher/src/production.ts
```diff
@@ -0,0 +1,88 @@
...
+import {
+	loadOldestUnpublishedAgentCoreDispatchAdmittedAt,
+	MAX_AGENTCORE_DISPATCH_PUBLISH_BATCH_SIZE,
+} from "@mymemo/agent-db/agentcore-dispatch";
+import type { Database } from "@mymemo/agent-db/client";
+import {
+	createDatabaseCanaryDispatchPublisherStore,
+	createSqsCanaryDispatchQueue,
+	createSsmCanaryEnablementControl,
+} from "agentcore-canary-dispatch/aws-adapters";
```
chengchao: [P3] Split publisher adapters from consumer-only adapters

The mixed `aws-adapters` module also statically imports the Bedrock AgentCore consumer client, and its workspace owns Secrets Manager dependencies. As a result, the independently filtered publisher image installs and loads consumer-only code even though its role intentionally has publisher-only authority. Split the SSM/SQS/store publisher adapters from the consumer/acquisition adapters to keep the image and change surface aligned with this service.

chengchao (commented):
Code review: 1 merge-blocking spec finding and 2 standards findings. The branch is otherwise green and mergeable.

apps/agentcore-dispatch-publisher/src/publisher-metrics.ts
chengchao: Addressed in ce8a40f. The production drain now carries the sampled pending age through a `PublisherTickFailure`, and the outer loop records that value on publication failure. If loading the oldest pending admission fails before age can be sampled, `PendingAgeMs` is omitted instead of reported as zero. Focused telemetry/loop tests, all eight workspace suites, and the rebuilt publisher image pass.

apps/agentcore-dispatch-publisher/package.json
chengchao: Addressed in 65418e6. Reusable publication behavior and adapters now live in `packages/agentcore-dispatch`; the dedicated app depends on that package and its Dockerfile no longer copies `apps/agentcore-canary-dispatch`. The image smoke check explicitly rejects the Canary workspace, Bedrock AgentCore client, and Secrets Manager client.

apps/agentcore-dispatch-publisher/src/production.ts
chengchao: Addressed in 65418e6. Publisher adapters are separate package entrypoints (`database-store`, `sqs-queue`, and `ssm-control`), while the Canary app's AWS adapter now contains only the Bedrock Runtime consumer. The rebuilt publisher image passes the dependency smoke check with Canary, Bedrock AgentCore, and Secrets Manager absent.

docs/adr/0026-publish-agentcore-dispatch-through-one-advisory-locked-loop.md
```diff
@@ -44,10 +44,10 @@ re-insertion, to shrink a table that stays at pending size anyway.
 
 ## Consequences
 
-- The compute home is constrained only to a long-lived process. The
-  agent-worker process, which outlives the rollout and already runs
-  advisory-locked maintenance loops, is the natural first home; moving the
-  loop later does not change this contract.
+- The publisher runs as one dedicated long-lived ECS task using the dedicated
```
chengchao: [P1] Do not rewrite the accepted publisher home without superseding the spec

Issue #481 says the publisher is a continuously running loop in the agent-worker process and requires it to start and drain with the worker. Parent #476 repeats that home and assigns publisher authority to the agent-worker task role. A dedicated ECS task starts and stops independently and moves that authority elsewhere; rewriting this accepted ADR consequence does not provide the missing product/architecture decision. Either keep the loop in agent-worker, or first revise/supersede the issues and record the new deployment decision explicitly. Until then this PR implements the opposite deployment boundary from its originating spec.

.github/workflows/agentcore-dispatch-publisher-image.yml
```diff
@@ -0,0 +1,31 @@
+name: AgentCore dispatch publisher image
+
+on:
+  pull_request:
+    paths:
+      - apps/agentcore-dispatch-publisher/**
```
chengchao: [P2] Trigger the image gate for every copied input

This path filter omits .dockerignore plus apps/*/package.json and packages/*/package.json, even though the Dockerfile copies every workspace manifest before the filtered install and .dockerignore controls the build context. A PR can therefore break the production image while this pre-merge image check is skipped. Mirror the established worker/runtime image inputs by adding those three patterns.

infra/terraform/iam.tf
```diff
@@ -58,6 +84,40 @@ resource "aws_iam_role" "agent_worker_task" {
...
+
+data "aws_iam_policy_document" "agentcore_dispatch_publisher" {
+  statement {
+    actions   = ["ssm:GetParameter"]
+    resources = [local.agentcore_dispatch_enabled_parameter_arn]
+  }
+
+  statement {
+    actions = [
+      "sqs:GetQueueAttributes",
```
chengchao: [P3] Remove the unused queue permission

The publisher only constructs SendMessageCommand; neither the app nor @mymemo/agentcore-dispatch reads queue attributes. ADR-0025 requires a per-workload least-privilege publisher role, so sqs:GetQueueAttributes is extra authority. Remove it unless a concrete runtime call is added and documented.

chengchao (commented):
Code review: 1 merge-blocking spec finding and 2 standards findings. Please resolve the P1 before merge.

packages/agentcore-dispatch/src/publisher.ts
```diff
@@ -50,9 +48,9 @@ export function createCanaryDispatchPublisher(options: {
 			const claimed = await options.store.claim({
 				publisherId: options.publisherId,
 				runId: input.runId,
-				limit: CANARY_QUEUE_INVARIANTS.publisherBatchSize,
+				limit: MAX_AGENTCORE_DISPATCH_PUBLISH_BATCH_SIZE,
```
chengchao: [P1] Stop before starting the next unit after shutdown

#481 requires SIGINT/SIGTERM to stop new work while allowing only the active send/confirm unit to settle. The process `AbortSignal` never reaches the per-row loop below, so a signal during row 1 still begins SSM/send/confirm for rows 2–10 in this claimed batch. Pass shutdown state into the publisher and check it between dispatches; leave the remainder leased for normal retry.

packages/agentcore-dispatch/src/sqs-queue.ts
```diff
@@ -0,0 +1,28 @@
...
+	): Promise<Pick<SendMessageCommandOutput, "MessageId">>;
+}
+
+export function createSqsAgentCoreDispatchQueue(options: {
+	client: SqsCommandClient;
+	queueUrl: string;
+}): AgentCoreDispatchQueue {
+	return {
+		async send(dispatch): Promise<void> {
+			await options.client.send(
```
chengchao: [P1] Bound AWS operations below the ECS stop timeout

This send, and the SSM send in `ssm-control.ts`, has neither a per-operation abort deadline nor a bounded request handler. A stalled AWS call can therefore outlive the task's 30-second `stopTimeout` and force ECS to kill the active unit, contrary to #481/ADR-0027. Give both operations explicit deadlines comfortably below 30 seconds.

apps/agentcore-dispatch-publisher/src/config.ts
```diff
@@ -0,0 +1,78 @@
...
+		`${name} must be a positive integer (got: ${raw})`,
+	);
+	return value;
+}
+
+/** Load only the database and AWS dispatch authority owned by this app. */
+export function loadAgentCoreDispatchPublisherConfigFromEnv(
+	env: Env,
+): AgentCoreDispatchPublisherConfig {
+	assert(env.AGENT_DATABASE_URL, "AGENT_DATABASE_URL is required");
```
chengchao: [P2] Reject malformed bootstrap configuration

These checks reject only missing values. Truthy junk such as a malformed database URL, whitespace region/parameter name, or non-SQS queue URL reaches the loop and becomes a recurring "transient" tick failure instead of terminating the process. #481 says invalid bootstrap configuration exits, so parse and validate these static values before constructing the clients.

apps/agentcore-dispatch-publisher/src/publisher-metrics.ts
```diff
@@ -0,0 +1,79 @@
...
+			CloudWatchMetrics: [
+				{ Namespace: namespace, Dimensions: [[]], Metrics: metrics },
+			],
+		},
+	};
+}
+
+export function recordPublisherLostLock(logger: PublisherLogger): void {
+	logger.info({
+		...publisherMetricEnvelope([{ Name: "PublisherLostLock", Unit: "Count" }]),
```
chengchao: [P2] Emit `lock_not_acquired`, not a lost-lock event

#481 and ADR-0027 define normal overlap contention as informational `lock_not_acquired` telemetry and explicitly note that the task did not lose a lock it held. `PublisherLostLock` / `lost_lock` states the opposite and will not provide the event name expected by the follow-on observability work. Rename the metric, outcome, function, and test.

apps/agentcore-dispatch-publisher/src/publisher-loop.ts
```diff
@@ -0,0 +1,70 @@
...
+	intervalMs: number;
+	signal: AbortSignal;
+	wait?: (intervalMs: number, signal: AbortSignal) => Promise<void>;
+}
+
+/** Publish one batch while this task owns the deployment-overlap lock. */
+export async function publishAgentCoreDispatchTick(
+	options: PublisherTickOptions,
+): Promise<void> {
+	const locked = await tryWithAdvisoryLock(
```
chengchao: [P2] Check the fail-closed gate before taking the lock

#481 specifies the tick order as SSM gate check → advisory lock → drain. Here the lock is taken first; the pending-age query and SSM call happen inside the callback. A disabled or stalled SSM check therefore occupies the publishing critical section, and an overlapping task never performs its own gate check. Check the gate at the tick boundary before locking, while retaining the per-row recheck if desired.

chengchao (commented):
Latest-head review of `cf3f8c3`: five new actionable findings (2 P1, 3 P2). The two P1 shutdown findings are merge-blocking. The two existing unresolved findings about the publisher-image path filter and unused `sqs:GetQueueAttributes` permission also remain valid and are not duplicated here. The older publisher-home P1 is obsolete because #481/#476 and ADR-0027 now explicitly select the dedicated service. CI is green, but the shutdown acceptance criteria are not yet met.

docs/adr/0026-publish-agentcore-dispatch-through-one-advisory-locked-loop.md
chengchao: Resolved by the architecture decision now on `main`: ADR-0027 explicitly supersedes ADR-0026's initial compute-home consequence and selects the dedicated publisher service. The latest-head review also confirms this older finding is obsolete after #481/#476 and ADR-0027 were updated.

.github/workflows/agentcore-dispatch-publisher-image.yml
chengchao: Addressed in `2136999`. The publisher-image workflow now watches `.dockerignore`, `apps/*/package.json`, and `packages/*/package.json`, covering every manifest/build-context input copied by the Dockerfile.

infra/terraform/iam.tf
chengchao: Addressed in `2136999`. The dedicated publisher task policy now grants only `sqs:SendMessage` on the queue; the unused `sqs:GetQueueAttributes` permission was removed. Terraform formatting and validation pass.

packages/agentcore-dispatch/src/publisher.ts
chengchao: Addressed in `2136999`. The process `AbortSignal` now reaches the shared publisher and is checked between claimed dispatches and again after each per-row gate check. An active send still reaches confirm, while later rows remain leased and no new unit starts; the focused unit test covers exactly that sequence.

packages/agentcore-dispatch/src/sqs-queue.ts
chengchao: Addressed in `2136999`. Both SQS `SendMessage` and SSM `GetParameter` now receive a fixed 10-second abort deadline, comfortably below the ECS 30-second `stopTimeout`; adapter tests verify an abort signal is passed to each AWS request.

apps/agentcore-dispatch-publisher/src/publisher-metrics.ts
chengchao: Addressed in `2136999`. The function, outcome, EMF metric, value field, and tests now consistently use `lock_not_acquired` / `PublisherLockNotAcquired`; normal deployment overlap is logged as informational contention.

apps/agentcore-dispatch-publisher/src/config.ts
chengchao: Addressed in `2136999`. Startup now rejects malformed PostgreSQL URLs, invalid/whitespace AWS regions, non-SQS or cross-region queue URLs, and invalid/whitespace SSM parameter names before constructing clients. Focused config tests cover the malformed cases.

apps/agentcore-dispatch-publisher/src/publisher-loop.ts
chengchao: Addressed in `2136999`. Every tick now checks the fail-closed SSM gate before opening a Postgres lock connection, rechecks shutdown before locking, and retains the existing in-lock/per-row gate checks. The focused test proves a disabled gate never opens the lock connection.

packages/agentcore-dispatch/src/publisher.ts
```diff
@@ -50,17 +49,19 @@ export function createCanaryDispatchPublisher(options: {
 			const claimed = await options.store.claim({
```
chengchao: [P1] Recheck shutdown before claiming the batch

Issue #481 requires SIGINT/SIGTERM to stop new ticks and batches. If shutdown arrives while the initial `control.isEnabled()` await is in flight and that read resolves `true`, this immediately claims a fresh batch; the per-row abort check then returns without sending, leaving up to 10 rows behind the three-minute claim lease. A focused regression test that aborts inside the gate read observes one claim instead of zero. Recheck `signal.aborted` after the gate read and before `store.claim()`, and cover that boundary in the publisher test.

apps/agentcore-dispatch-publisher/src/config.ts
```diff
@@ -0,0 +1,147 @@
...
+			: `sqs.${region}.amazonaws.com`;
+		assert(
+			url.protocol === "https:" &&
+				url.hostname === hostname &&
+				url.port === "" &&
+				url.username === "" &&
+				url.password === "" &&
+				url.search === "" &&
+				url.hash === "" &&
+				url.pathname.split("/").filter(Boolean).length >= 2,
```
chengchao: [P2] Reject malformed QueueUrl paths at bootstrap

The `>= 2` check accepts values such as `https://sqs.us-west-2.amazonaws.com/not-an-account/dispatch/extra`. That is not an SQS QueueUrl, but the process starts and every tick then treats the AWS rejection as a transient failure. #481 says invalid bootstrap configuration exits, so require exactly a 12-digit account segment plus a valid standard-queue name, with no extra segments, and add those malformed cases to `config.test.ts`.

docs/agents/agent-worker.md
```diff
@@ -42,6 +42,12 @@ A qualifying `done` or `interrupted` terminal transaction publishes `conversatio
 
 `src/index.ts` resolves and verifies the SDK CLI before claiming work, creates the per-Conversation Fargate query directory, and wires the real SDK processor. When a Run ends with an `error` Outcome, persist only a generic client-facing message and keep internal details in trusted-worker logs.
 
+## Related AgentCore dispatch publisher
+
+`apps/agentcore-dispatch-publisher/src/main.ts` is the dedicated app entrypoint for the one publisher ECS task from ADR-0026. The Conversation-serving `apps/agent-worker/src/index.ts` does not start it and does not read its SQS or SSM configuration. The publisher task checks the fail-closed SSM gate, claims a bounded outbox batch, sends strict envelopes to SQS, and confirms successful sends. Its tick-scoped session advisory lock only excludes the old/new task overlap during a rolling deployment.
```
chengchao: [P3] Attribute the dedicated service to ADR-0027

ADR-0026 originally placed publication in `agent-worker`; ADR-0027 explicitly supersedes that compute-home consequence and selects this dedicated ECS service. Point this sentence at ADR-0027 so the operational guide does not attribute the new deployment boundary to the decision that chose the old one.

docs/agents/agent-worker.md
```diff
@@ -42,6 +42,12 @@ A qualifying `done` or `interrupted` terminal transaction publishes `conversatio
...
+
+`SIGINT` or `SIGTERM` aborts the sleep after a tick; shutdown waits for an active tick to release its database connection and then closes the pool. The loop emits CloudWatch embedded metrics for lost locks, errors, and the oldest unpublished admission age. It does not publish from Run admission and does not delete outbox audit rows.
```
chengchao: [P3] Describe contention as lock_not_acquired

This task never held and lost the lock. Issue #481 and ADR-0027 define normal rollout contention as informational `lock_not_acquired`, and the implementation emits `PublisherLockNotAcquired`. Replace `lost locks` with that terminology so the guide does not mislead the follow-on alarm wiring in #483.

chengchao (commented):
Latest-head two-axis review: 2 merge-blocking Spec findings (P1/P2) and 2 documentation findings. Focused publisher/deployment tests otherwise pass, and all GitHub checks are green.

packages/agentcore-dispatch/src/publisher.ts
chengchao: Addressed in `64f3f7b`. `publishPending()` now rechecks the process `AbortSignal` immediately after the initial gate read and before `store.claim()`. The focused regression test aborts inside that gate read and verifies that no claim, send, or confirm begins.

apps/agentcore-dispatch-publisher/src/config.ts
chengchao: Addressed in `64f3f7b`. Bootstrap validation now requires an exact `/<12-digit account>/<standard queue name>` path, with a 1–80 character alphanumeric/underscore/hyphen queue name and no extra segments. Tests cover a non-account segment, an extra segment, and a FIFO name.

docs/agents/agent-worker.md
chengchao: Addressed in `64f3f7b`. The worker operational guide now attributes the dedicated publisher ECS task to ADR-0027, which supersedes ADR-0026’s original compute-home consequence.

docs/agents/agent-worker.md
chengchao: Addressed in `64f3f7b`. The guide now describes normal rollout contention as `lock_not_acquired`, matching ADR-0027 and the emitted `PublisherLockNotAcquired` metric.

chengchao (commented):
Latest-head two-axis review: 1 merge-blocking finding shared by the Spec and Standards axes. All checks are green, but disabled-gate ticks still suppress the pending-age metric that #481 makes the primary paging symptom. The transitional publisher Lambda and repair rule are intentionally removed by follow-up #483 and are not a #491 finding.

apps/agentcore-dispatch-publisher/src/publisher-loop.ts
```diff
@@ -0,0 +1,74 @@
...
+	intervalMs: number;
+	signal: AbortSignal;
+	wait?: (intervalMs: number, signal: AbortSignal) => Promise<void>;
+}
+
+/** Publish one batch while this task owns the deployment-overlap lock. */
+export async function publishAgentCoreDispatchTick(
+	options: PublisherTickOptions,
+): Promise<void> {
+	if (options.signal?.aborted || !(await options.publisher.isEnabled())) return;
```
chengchao: [P1] Preserve pending-age telemetry while disabled

Issue #481 makes pending age the primary paging symptom and requires publisher emission to include it. Returning here means a persistently disabled gate never reaches loadOldestAdmittedAt or recordPublisherPublication, so backlog can grow while PendingAgeMs disappears entirely. The recorder unit test for a disabled result does not exercise this production path. Keep publication fail-closed, but sample and emit pending age for disabled ticks and cover the behavior through publishAgentCoreDispatchTick.

Standards: this preflight also duplicates the initial enablement check in packages/agentcore-dispatch/src/publisher.ts; splitting that decision across two layers is what lets telemetry behavior drift. Consolidate the check or explicitly carry its result into the age-sampling path.

apps/agentcore-dispatch-publisher/src/advisory-lock.ts
```diff
@@ -0,0 +1,34 @@
...
+	release(): void;
+}
+
+/** Run a callback while one dedicated pool connection owns a session lock. */
+export async function tryWithAdvisoryLock<T>(
+	pool: AdvisoryLockPool,
+	key: number,
+	callback: () => Promise<T>,
+): Promise<{ ran: true; result: T } | { ran: false }> {
+	const client = await pool.connect();
```
chengchao: [P1] Handle lock-connection errors as recoverable tick failures

Issue #481 requires transient Postgres tick failures to keep the loop alive and emit error telemetry. `pg-pool` removes its idle `error` listener while this client is checked out, and `pg` emits `error` when the live backend/socket fails. Because this checkout installs no listener, that event is uncaught and terminates Bun before `runAgentCoreDispatchPublisher` can record `PublisherErrors` or continue; the real-Postgres test masks the production path by installing its own client listener. Add a scoped listener, stop further batches/rows if the lock session dies while allowing the active send/confirm unit to settle, reject the tick into the loop's telemetry path, and discard the broken client.

apps/agentcore-dispatch-publisher/src/config.ts
```diff
@@ -0,0 +1,148 @@
...
+	const match = /^([a-z]+:\/\/)([^@/]+)@(.*)$/i.exec(url);
+	if (!match) return url;
+	const [, scheme, userinfo, rest] = match;
+	if (!scheme || !userinfo || rest === undefined || userinfo.includes(":")) {
+		return url;
+	}
+	return `${scheme}${userinfo}:${encodeURIComponent(password)}@${rest}`;
+}
+
+function withSsl(url: string, enabled: boolean): string {
```
chengchao: [P3] Share the database URL parsing contract

This `withPassword`/`withSsl` logic duplicates `apps/agent-worker/src/config/env.ts` and `apps/chat-api/src/config/env.ts`, while the publisher configuration guide explicitly says `DB_PASSWORD` and `DB_SSL` have the same behavior as the worker. This is a judgment-call Duplicated Code smell rather than a merge blocker, but keeping the TLS/password semantics in three copies invites security/configuration drift. Consider extracting the shared DB URL resolution primitives while leaving app-specific validation local.

chengchao (commented):
Latest-head two-axis review: 1 merge-blocking Spec finding (P1) and 1 non-blocking Standards smell (P3). All current GitHub checks are green, but the unchecked advisory-lock client error path can terminate the publisher instead of emitting failure telemetry and continuing.

apps/agentcore-dispatch-publisher/src/advisory-lock.ts
chengchao: Addressed in `c2c53ba`. The checked-out lock client now has a scoped `error` listener that aborts a lock-session signal, waits for the callback’s active unit to settle, then rejects the tick and releases the client with the error so `pg-pool` discards it. Production combines that signal with process shutdown, so later batches/rows stop while an active send/confirm can finish. The loop test verifies `PublisherErrors` emission and a successful next tick on a fresh connection; the real-Postgres test no longer installs its own masking listener and now waits on the production lock signal.

apps/agentcore-dispatch-publisher/src/config.ts
chengchao: Addressed in `c2c53ba`. Password splicing and the shared `DB_SSL` policy now live in `@mymemo/agent-db/database-url`; publisher, agent-worker, and chat-api all use that resolver, while their app-specific required-field and URL validation remains local. Focused resolver/config tests and all three app typechecks pass.

apps/agentcore-dispatch-publisher/src/advisory-lock.ts
```diff
@@ -0,0 +1,68 @@
...
+		}
+		if (clientError) throw clientError;
+		if (acquired.rows[0]?.locked !== true) return { ran: false };
+
+		let result: T;
+		try {
+			result = await callback(lockSession.signal);
+		} catch (error) {
+			if (clientError) throw clientError;
+			throw error;
```
chengchao: [P1] Release or destroy the session after callback failure

When callback rejects for a transient tick failure, control jumps past pg_advisory_unlock and finally calls client.release() without an error. node-postgres returns that still-healthy session to the pool while it still owns the advisory lock; later acquisitions on the same session are reentrant, and other tasks/sessions can remain locked out beyond this tick. This violates the tick-scoped lock contract. Please unlock in a finally around the callback, or destroy the connection on every exceptional callback exit.

packages/agentcore-dispatch/src/publisher.ts
```diff
@@ -14,53 +16,59 @@ export interface CanaryDispatchPublisherStore {
...
 			};
 			for (const dispatch of claimed) {
+				if (options.signal?.aborted) return result;
 				if (!(await options.control.isEnabled())) {
 					return { ...result, status: "disabled" };
 				}
+				if (options.signal?.aborted) return result;
 				try {
 					await options.queue.send(dispatch);
 					const confirmed = await options.store.confirm({
```
chengchao: [P1] Do not confirm after losing the advisory-lock session

The production lock-loss signal is checked before queue.send, but not after it. If the Postgres lock session dies while the SQS request is in flight, another task can acquire the advisory lock; this task can then return from send and still execute confirm concurrently with the new critical section. Keep graceful-shutdown settling separate from lock loss, and abort/treat this unit as ambiguous when the lock signal fires before confirmation.

chengchao (commented):
Two merge-blocking advisory-lock lifecycle findings.

apps/agentcore-dispatch-publisher/src/advisory-lock.ts
chengchao: Addressed in `d9d3f0d`. `tryWithAdvisoryLock` now captures the callback outcome, attempts `pg_advisory_unlock` before propagating any callback failure, and only then returns the session to the pool. A client or unlock failure is still passed to `release(error)` so the session is discarded. The regression test proves a rejected callback issues the unlock query and releases the now-unlocked client.

packages/agentcore-dispatch/src/publisher.ts
chengchao: Addressed in `d9d3f0d`. Graceful shutdown and lock loss are now separate signals. Shutdown still allows the active send/confirm unit to settle, but after `queue.send` the publisher checks the lock-session signal before confirmation; lock loss marks that Run ambiguous, skips `confirm`, and stops the batch. The regression test covers lock loss during send and verifies no confirmation or later send occurs.

packages/agent-db/src/database-url.ts
```diff
@@ -0,0 +1,41 @@
...
+	if (dbPassword) {
+		const match = /^([a-z]+:\/\/)([^@/]+)@(.*)$/i.exec(resolved);
+		if (match) {
+			const [, scheme, userinfo, rest] = match;
+			if (scheme && userinfo && rest !== undefined && !userinfo.includes(":")) {
+				resolved = `${scheme}${userinfo}:${encodeURIComponent(dbPassword)}@${rest}`;
+			}
+		}
+	}
+	if (dbSsl !== "disable" && !/[?&]sslmode=/.test(resolved)) {
```
chengchao: [P2] Enforce the documented local-only plaintext exception

`DB_SSL=disable` skips TLS for every hostname, even though this helper says callers may disable TLS only for local Postgres and `docs/agents/configuration.md` makes the same restriction authoritative. The new publisher therefore accepts a remote plaintext `AGENT_DATABASE_URL` as valid bootstrap configuration. Reject `disable` unless the parsed host is `localhost`, `127.0.0.1`, or `[::1]` (ideally here so every caller shares the enforcement).

infra/terraform/shared_state.tf
```diff
@@ -25,3 +25,11 @@ data "aws_ecs_cluster" "shared" {
 
   cluster_name = local.shared_ecs_cluster_name_output
 }
+
+data "aws_caller_identity" "current" {}
+
+data "aws_partition" "current" {}
+
+data "aws_kms_alias" "agentcore_dispatch_queue" {
```
chengchao: [P1] Pre-apply the deploy-role permission needed by this lookup

This data source needs `kms:ListAliases`/`kms:DescribeKey`, but the release workflow runs the production Terraform plan without applying `infra/bootstrap-iam`. A read-only check of the live `mymemo-agent-github-actions-deploy` inline policy confirms that it still lacks the new `AgentCoreDispatchKmsMetadata` statement, so the first post-merge coordinated release will fail at this lookup before it can deploy anything. #481 requires complete coordinated release wiring. Please pre-apply this bootstrap change and document that existing-environment prerequisite in this PR (as the Redis section already does for its bootstrap permissions), or remove the plan-time dependency on permissions the deploy role does not yet have.

apps/agentcore-dispatch-publisher/src/config.ts
```diff
@@ -0,0 +1,131 @@
...
+
+function positiveIntOr(
+	raw: string | undefined,
+	fallback: number,
+	name: string,
+): number {
+	if (raw === undefined) return fallback;
+	const value = Number(raw);
+	assert(
+		Number.isInteger(value) && value > 0,
```
chengchao: [P2] Reject intervals that overflow Bun timers

This accepts any positive integer, but Bun 1.3 converts values above `2_147_483_647` to a 1 ms timeout (`TimeoutOverflowWarning`). A malformed interval such as `2147483648` therefore passes bootstrap and makes the publisher hammer SSM/Postgres instead of waiting, contrary to Issue #481, which requires invalid bootstrap configuration to exit. Bound this value to the Bun timer range here and in the matching Terraform variable validation.

packages/agent-db/src/database-url.ts
chengchao: Verified and acknowledged. Production keeps `DB_SSL=require`, so this PR does not introduce plaintext database traffic; the remote-host hardening is outside this PR’s regression scope. No code change was made for this withdrawn finding.

infra/terraform/shared_state.tf
chengchao: Verified and acknowledged. The bootstrap IAM apply completed with 0 added, 1 changed, and 0 destroyed, and a follow-up production bootstrap plan reports no changes. The release role now has the required KMS metadata permissions, so no further PR change is needed for this withdrawn blocker.

apps/agentcore-dispatch-publisher/src/config.ts
chengchao: Addressed in `3f1ab83`. Runtime config now rejects intervals above `2_147_483_647`, and the Terraform variable enforces the same upper bound. Regression coverage accepts the exact Bun timer maximum and rejects `2147483648`; publisher tests, scoped typecheck, Terraform validation, deployment contracts, and the full workspace suite pass.

apps/agentcore-dispatch-publisher/package.json
```diff
@@ -0,0 +1,26 @@
...
+	"name": "agentcore-dispatch-publisher",
+	"scripts": {
+		"test": "bun test --timeout=30000",
+		"start": "bun run src/main.ts",
+		"format": "biome format --write",
+		"lint": "biome lint --write",
+		"check": "biome check --write"
+	},
+	"dependencies": {
+		"@aws-sdk/client-sqs": "^3.1110.0",
```
chengchao: [P1] Keep publisher dependencies out of the agent-worker image

Issue #481 requires `agent-worker` to contain no Dispatch-publisher dependencies, but `apps/agent-worker/Dockerfile` still runs an unfiltered root `bun install --production` and then copies the entire root `node_modules` into the worker image. These publisher-only SQS/SSM clients are therefore shipped in the Conversation executor even though its package/import graph and IAM no longer use them. Filter the worker install to the `agent-worker` workspace and extend `agent-worker-image-check.sh` to assert that the dispatch-only clients/package are absent.

chengchao (commented):
Reviewed the current head against issue #481 and the repository/ADR standards. One spec-blocking dependency-boundary finding remains; details are inline. The standards pass found no additional actionable current-head violations.

apps/agentcore-dispatch-publisher/package.json
chengchao: Addressed in `299114d`. The worker image now runs Bun's production install with `--filter agent-worker`, limiting `node_modules` to the worker's transitive workspace graph. The strengthened image smoke contract asserts that `@aws-sdk/client-sqs`, `@aws-sdk/client-ssm`, `@mymemo/agentcore-dispatch`, and `agentcore-dispatch-publisher` are absent while still executing the bundled Claude SDK CLI. A real linux/amd64 image build completed with 200 installed packages and passed the smoke check; `bun test scripts/deploy-config.test.ts`, `bash -n`, and `git diff --check` also pass.

apps/agentcore-dispatch-publisher/src/config.ts
```diff
@@ -0,0 +1,132 @@
...
+	} catch {
+		throw new Error(
+			"CANARY_DISPATCH_QUEUE_URL must be an HTTPS SQS URL in AWS_REGION",
+		);
+	}
+}
+
+function validateParameterName(value: string): string {
+	assert(
+		/^[A-Za-z0-9_.\-/]+$/.test(value),
```
chengchao: [P2] Reject invalid SSM names during bootstrap

Issue #481 requires invalid bootstrap configuration to end the process, but this character-only regex accepts names such as `mymemo/dispatch/enabled` (a hierarchy without the required leading `/`), reserved `/aws/...` or `/ssm/...` prefixes, and hierarchies deeper than 15 levels. AWS rejects those in `GetParameter`; the process then treats the bad static value as a transient tick failure and retries forever. Validate the full Parameter Store naming contract here (simple name or leading-slash hierarchy, allowed prefix, depth, and length) and add rejection tests.

apps/agentcore-dispatch-publisher/src/config.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function positiveIntOr(
+	raw: string | undefined,
+	fallback: number,
+	name: string,
+): number {
+	if (raw === undefined) return fallback;
+	const value = Number(raw);
+	assert(
+		Number.isInteger(value) && value > 0 && value <= MAX_TIMER_INTERVAL_MS,
```
chengchao: [P2] Keep the configured tick interval seconds-scale

ADR-0026 calls for a couple-second tick and Issue #481 requires seconds-scale healthy publication latency, but this accepts `2_147_483_647` ms—about 24.9 days. That value no longer overflows Bun, yet it leaves a healthy publisher asleep long enough to violate the core latency contract. Cap the runtime and matching Terraform variable to a documented seconds-scale operational range (or remove the override if only the 2-second cadence is supported).

chengchao (commented):
Updated review: both configuration findings are withdrawn. They concern trusted, Terraform-owned operator settings with valid checked-in defaults, not defects in this PR. No actionable findings remain. All CI and focused verification passed. Merge recommendation: yes.

apps/agentcore-dispatch-publisher/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+FROM base AS install
+COPY --from=manifests /usr/src/app /temp/prod
+RUN cd /temp/prod \
+ && bun install --frozen-lockfile --production --filter agentcore-dispatch-publisher
+
+FROM base AS release
+COPY --from=install /temp/prod/node_modules node_modules
+COPY apps/agentcore-dispatch-publisher apps/agentcore-dispatch-publisher
+COPY packages/agent-db packages/agent-db
```
chengchao: [P1] Keep opposite-side Dispatch code out of both runtime images

This copies the entire `@mymemo/agent-db` source package into the publisher image. Its `src/agentcore-dispatch.ts` module combines publisher claim/confirm operations with consumer Durable acquisition and Conversation Ownership transitions, so Bun loads and exposes consumer-side code in the publisher process. Conversely, `apps/agent-worker/Dockerfile` copies the same broad package, leaving publication claim/confirm code in the worker image. Issue #481 explicitly requires the filtered publisher image to exclude consumer application code and `agent-worker` to contain no Dispatch publication code. Split the publisher DB operations from acquisition/Run-serving modules, then bundle or copy only each runtime's reachable side.

chengchao (commented):
Latest-head review: one merge-blocking Spec finding remains. The focused runtime/deployment verification and all current GitHub checks are green, but the source-level process/image boundary required by issue #481 is not yet established.

apps/agentcore-dispatch-publisher/Dockerfile
chengchao: Withdrawn — no merge issue

I overinterpreted issue #481's process and dependency boundary as a requirement for physical source containment. `@mymemo/agent-db` intentionally owns the shared writable Postgres state used by both Dispatch publication and Durable acquisition. The required runtime separation is present: only the Dispatch publisher owns the publication lifecycle and SSM/SQS/queue-KMS authority, while `agent-worker` has neither. Splitting the mixed database module could be an optional organization improvement, but it is not required for this PR. No code change is needed, and this finding does not block merge.

chengchao (commented):
Updated latest-head review: the source-containment P1 is withdrawn and its thread is resolved. The shared `@mymemo/agent-db` package is consistent with the intended Postgres trust model; the required Dispatch publisher versus `agent-worker` lifecycle and AWS-authority separation is present. No actionable findings remain. All six checks pass, GitHub reports the PR clean and mergeable, and my merge recommendation is yes.

chengchao (commented):
Reviewed pinned head `299114d` against issue #481, ADR-0026/ADR-0027, and the repository standards.

## Standards

No actionable findings.

## Spec

No actionable findings. The dedicated publisher service, advisory-lock behavior, fail-closed gate, bounded multi-batch drain, at-least-once recovery, shutdown handling, metric emission, filtered runtime boundaries, IAM/networking, and release wiring match #481. Alarm wiring and retirement of the dormant canary publisher/repair path remain explicitly owned by #483.

Summary: Standards 0 findings; Spec 0 findings. All six current checks pass, GitHub reports the PR clean and mergeable, and the merge recommendation is yes.

PR #493, opened 2026-08-19
feat(agentcore): harden production dispatch posture

### LINKED ISSUE #482: AgentCore rollout: consumer pre-check and production config posture
## Parent

#476

## What to build

The remaining production posture in code. The consumer pre-checks Run status in Postgres before invoking, so an already-terminal Run (including a queued Run interrupted before delivery) acknowledges for one indexed read instead of a Runtime invocation. The SSM parameter and every env var shed their canary names for dispatch vocabulary. The Runtime publishes Downloadable artifacts through the standard production namespace — the canary prefix disappears from code — so both runtimes' artifact keys are indistinguishable to chat-api's signer. The shared infra invariants are retuned for production: visibility timeout above the consumer timeout but production-low, a raised max receive count so the Postgres queued timeout (not SQS receive counting) is the effective backstop, batch size one, no reserved concurrency.

## Acceptance criteria

- [x] The consumer acknowledges a terminal Run without invoking; non-terminal Runs proceed to invocation with the matrix otherwise unchanged.
- [x] No canary-named env vars or SSM paths remain in code; config validation reflects the new names.
- [x] Runtime artifact keys use the standard namespace; no canary prefix constant remains.
- [x] The invariants module encodes the retuned values and its test enforces the relationships between them.

## Blocked by

- #478 (production dispatch contract)

chengchao:
## Summary

- Pre-check durable Run status before AgentCore invocation and acknowledge terminal deliveries without invoking the Runtime.
- Replace canary-named application environment variables and SSM paths with production AgentCore dispatch vocabulary.
- Publish Downloadable artifacts through the standard `objects/` namespace and align Runtime IAM.
- Apply production queue, concurrency, metric, and alarm invariants to both code and Terraform.
- Run Redis-backed tests against GitHub Actions service containers while preserving disposable local Redis fixtures.

## Acceptance criteria

- [x] Terminal Runs are acknowledged without Runtime invocation; the non-terminal disposition matrix is unchanged.
- [x] Application environment variables and SSM paths use production AgentCore dispatch names with fail-closed validation.
- [x] Runtime artifacts use the standard `objects/` namespace with no canary prefix constant.
- [x] Queue invariants encode and deploy visibility `180`, max receive count `5`, batch size `1`, and no reserved concurrency; tests enforce their relationships.

## Verification

- `bun install --frozen-lockfile`
- `bun run test` — 1,130 passed; 27 environment-gated Postgres tests skipped locally
- Redis relay contract, failure matrix, and AgentCore execution-service tests against `TEST_REDIS_URL`
- Redis relay contract with the disposable-local fallback
- Scoped Biome checks and `git diff --check`
- `terraform -chdir=infra/agentcore-canary validate`
- Two-axis final review: no Standards or Spec findings
- All four PR review threads resolved

Closes #482


packages/agent-db/src/agentcore-dispatch.ts
```diff
@@ -85,6 +89,39 @@ export async function loadOldestUnpublishedAgentCoreDispatchAdmittedAt(
...
+			eq(agentCoreDispatchOutbox.runId, runs.runId),
+		)
+		.where(
+			and(
+				eq(runs.runId, dispatch.runId),
+				eq(runs.userId, dispatch.userId),
+				eq(runs.conversationId, dispatch.conversationId),
+				eq(agentCoreDispatchOutbox.userId, dispatch.userId),
+				eq(agentCoreDispatchOutbox.conversationId, dispatch.conversationId),
+				eq(agentCoreDispatchOutbox.admittedAt, dispatch.admittedAt),
```
chengchao: [P1] Avoid exact SQL equality after the timestamp crosses a JS `Date`

Production admission omits `admittedAt`, so PostgreSQL's `default now()` can retain microseconds. The publisher reads that value into a JS `Date` and the envelope serializes it at millisecond precision; binding the truncated value back into this exact `timestamptz` equality will therefore usually miss the outbox row. A late/duplicate delivery for a terminal Run then gets `null` here and invokes AgentCore, defeating issue #482's required precheck. Please load the exact row by indexed Run identity and compare the returned timestamp with `getTime()` (as the acquisition path already does), or otherwise normalize precision at storage, and cover the production default-timestamp path in a real-Postgres test.

chengchao (commented):
Found one merge-blocking Spec issue in the terminal-Run precheck. Standards review found no issues.

scripts/agentcore-canary-infra.test.ts
```diff
@@ -225,12 +225,7 @@ describe("AgentCore canary dispatch infrastructure", () => {
 		expect(source).toMatch(
 			/sid\s*=\s*"WriteSyntheticArtifactsOnly"[\s\S]*?resources\s*=\s*\["arn:aws:s3:::\$\{var\.artifact_bucket_name\}\/\$\{local\.canary_artifact_object_key_prefix\}\/\*"\]/,
 		);
-		expect(source).toMatch(
-			/canary_artifact_object_key_prefix\s*=\s*"objects\/agentcore-canary"[\s\S]*?CANARY_ARTIFACT_OBJECT_KEY_PREFIX\s*=\s*local\.canary_artifact_object_key_prefix/,
-		);
-		expect(canaryProduction).toContain(
-			"artifactObjectKeyPrefix: options.bootstrap.artifactObjectKeyPrefix",
-		);
+		expect(canaryProduction).not.toContain("artifactObjectKeyPrefix");
```
chengchao: [P1] Align Runtime artifact IAM with the standard object namespace. This new assertion removes the Runtime prefix override, so createArtifactPublisher emits objects/<uuid>, but the assertion immediately above still preserves the IAM resource backed by local.canary_artifact_object_key_prefix, which remains objects/agentcore-canary in locals.tf and iam.tf. S3 will deny every AgentCore artifact upload. Delete the canary prefix constant and scope both IAM and this test to objects/*.

apps/agentcore-canary-dispatch/src/invariants.ts
```diff
@@ -1,17 +1,18 @@
...
 	partialBatchResponses: true,
 	consumerTimeoutSeconds: 120,
-	visibilityTimeoutSeconds: 300,
+	visibilityTimeoutSeconds: 180,
+	queuedRunTimeoutSeconds: AGENTCORE_UNOWNED_QUEUE_TIMEOUT_MS / 1_000,
 	retentionSeconds: 24 * 60 * 60,
-	maxReceiveCount: 3,
-	reservedConsumerConcurrency: 1,
+	maxReceiveCount: 5,
+	reservedConsumerConcurrency: null,
```
chengchao: [P1] Apply these retuned invariants to the deployed Terraform. This module now declares visibility 180, maxReceiveCount 5, and no reserved concurrency, but infra/agentcore-canary/queue.tf still deploys 300 and 3, infra/agentcore-canary/lambdas.tf still reserves 1, and the infra test still asserts those old values. Issue #482 requires the production settings themselves, so changing only this TypeScript object leaves production unchanged.

apps/agentcore-canary-dispatch/src/production.ts
```diff
@@ -107,7 +113,7 @@ export function createEmbeddedMetricCanaryDispatchAlarm(
 						Timestamp: Date.now(),
 						CloudWatchMetrics: [
 							{
-								Namespace: "MyMemo/AgentCoreCanary",
+								Namespace: "MyMemo/AgentCoreDispatch",
```
chengchao: [P1] Keep emitted metrics and CloudWatch alarms in the same namespace. This change emits MyMemo/AgentCoreDispatch, while infra/agentcore-canary/alarms.tf and the deployment checker still watch MyMemo/AgentCoreCanary. PoisonDispatch and DisabledDelivery will therefore never trip the configured safety alarms. Migrate the alarm and inspection configuration in this PR, or retain the existing namespace.

chengchao (commented):
Current-head follow-up review: no findings.

- Standards: pass.
- Spec (#482): pass.
- The four earlier review issues (timestamp precision, artifact IAM namespace, Terraform invariant values, and CloudWatch metric namespace) are resolved at `8ee33c2`.
- Local full-suite verification passes, and GitHub's infrastructure, integration, and image checks pass. The general `test` check is still in progress.

Issue #595, opened 2026-08-27
Spec: run the AI SDK chat route on HarnessAgent in a Vercel Sandbox


chengchao:
## Problem Statement

The AI SDK chat route `POST /api/chat` is mounted only by the local chat-api composition and proxies each message to the Agent-query Runtime on AgentCore, which runs a no-tool Claude Agent SDK query and returns raw Claude SDK NDJSON. Continuity relies on a detached in-memory SessionStore persisted to S3. The route has no tools, no standard client wire format, and a second deployed runtime with its own image, IAM, Terraform, and `/invocations` hop for one verification path.

Vercel's `@ai-sdk/harness` gives the same loop — Claude Code as the agent runtime, a sandbox as its host, an AI SDK `Agent` interface on top — without MyMemo owning the runtime. This spec moves `/api/chat` onto it and retires the Agent-query Runtime. It supersedes [Spec: replace durable Runs with direct AI SDK AgentCore streaming](https://github.com/X-GPT/mymemo-agent/issues/560); the decisions were made on [Wayfinder map: HarnessAgent on Vercel Sandbox for /api/chat](https://github.com/X-GPT/mymemo-agent/issues/585) and are recorded in [ADR-0033](../blob/main/docs/adr/0033-host-the-ai-sdk-chat-loop-in-a-vercel-sandbox-through-harnessagent.md).

## Solution

chat-api hosts `HarnessAgent` in-process. For each message it authorizes the Conversation exactly as today, then resumes (or creates) the Conversation's **Harness sandbox** — one persistent Vercel Sandbox named from the Conversation id — runs one turn of Claude Code inside it through the upstream `createClaudeCode` adapter, streams the turn back as the AI SDK UI message stream, and stops the sandbox so Vercel snapshots it. The sandbox snapshot carries the Conversation's model-side memory across turns; chat-api keeps only a small resume pointer per Conversation.

Stage 1 uses Claude Code's built-in tools inside the sandbox. Stage 2 — disabling built-ins, exposing MyMemo's existing tool catalog as HarnessAgent user tools, and reattaching the E2B Workspace — is named here and designed later.

## User Stories

1. As a Conversation user, I want a second message to continue the Agent's context from my first, so that I do not repeat myself.
2. As a Conversation user, I want continuation to survive a chat-api restart, so that context does not depend on one process.
3. As a Conversation user, I want the response streamed as it is produced, so that persistence does not turn chat into a batch operation.
[...]

PR #604, opened 2026-08-27
feat(chat-api): continue a Conversation across Harness turns

### LINKED ISSUE #599: Continue a Conversation across Harness turns
## Parent

[Spec: run the AI SDK chat route on HarnessAgent in a Vercel Sandbox](https://github.com/X-GPT/mymemo-agent/issues/595)

## What to build

The second message in a Conversation remembers the first, and still does after chat-api restarts. Each Conversation owns one persistent Harness sandbox named from its id: after a turn the route stops the sandbox (Vercel snapshots it) and stores the harness's small resume pointer; the next turn resumes by name with that pointer. If the sandbox cannot be resumed (expired or gone), the route starts a fresh sandbox and a fresh thread for the same id and logs it — the user is never blocked.

## Acceptance criteria

- [x] A migration adds a nullable JSON `harness_resume_state` column to `conversation_runtime`, owned by the shared data package and covered by the PGlite harness.
- [x] After a completed (or stopped/failed) turn the route calls `stop()` instead of destroying the session and stores the returned state as the Conversation's pointer.
- [x] A turn with a stored pointer creates the session with `{ sessionId: <conversationId>, resumeFrom: <pointer> }`; without one it creates a fresh session for the same id.
- [x] When resuming throws, the route creates a fresh session for the same id, logs a structured event, and continues; a fake-agent test proves this yields a response rather than an error.
- [x] The pointer is never interpreted or exposed; it does not appear in the stream.
- [x] Vercel default retention is used (30-day snapshot expiry, one snapshot kept); no deletion hook or sweep is added.
- [x] Deleting a Conversation deletes its row, and with it the pointer; the sandbox is left to Vercel's expiry.
- [x] A documented local two-turn check (command in `docs/agents/chat-api.md` or `development.md`) proves recall through the real harness, including across a chat-api restart.

## Blocked by

- Stop and fail a Harness turn cleanly.

chengchao:
## Summary

Closes #599 (parent spec #595, ADR-0033).

Each Conversation now owns one persistent Harness sandbox named from its id:

- A turn loads the Conversation's opaque resume pointer and calls `createSession({ sessionId, resumeFrom })`; without one it creates a fresh session for the same id.
- After the stream is drained, cancelled, or fails, the route calls `session.stop()` (Vercel snapshots the sandbox) and stores the returned state as the pointer. The pointer is never interpreted and never appears in the stream.
- If resuming throws (snapshot expired, sandbox gone), the route logs `harness session resume failed; starting a fresh session` and starts a fresh session for the same id — the user gets a response on a new thread.
- Retention is Vercel's default (30-day expiry, one snapshot kept); no deletion hook or sweep.

## Changes

- `packages/agent-db`: nullable jsonb `conversation_runtime.harness_resume_state` (migration `0028_medical_jubilee`), covered by the PGlite migration test.
- `apps/chat-api`: `HarnessResumeStateStore` + `PostgresHarnessResumeStateStore` (unfenced upsert — this path has no Run or Ownership epoch), wired into `AppDeps`; route resume / stop / fallback; Permanent deletion nulls the pointer.
- Docs: `docs/agents/chat-api.md` describes continuity and retention and adds the local two-turn recall check across a `docker compose restart chat-api`.

## Deviation from the acceptance criteria

"Deleting a Conversation deletes its row, and with it the pointer": the `conversation_runtime` row deliberately survives Permanent deletion today so the agent-worker cleanup sweep can still find and kill the E2B sandbox (`cleanup.ts` "Remove `conversation_runtime` rows whose conversation no longer exists"; existing test "leaves external cleanup state recoverable"). Deleting the row here would break that. So deletion nulls `harness_resume_state` in the same transaction; the sweep later removes the row. Net effect on the pointer is the same.

## Verification

- `apps/chat-api`: `bun test` — 146 pass (route tests: pointer stored after `stop()`, resume passes the stored pointer, resume failure → 200 + warn + replaced pointer; store round-trip on PGlite; delete nulls the pointer).
- `packages/agent-db`: `bun test` — 228 pass (new column exists after migrations).
- `biome check .` clean.
- The real-harness two-turn check is documented but not run here (needs Vercel + OpenRouter credentials).

### DIFF
--- a/apps/chat-api/src/deps.ts
+++ b/apps/chat-api/src/deps.ts
@@ -8,6 +8,10 @@ import {
 import type { Env as PinoEnv } from "hono-pino";
 import type { ApiConfig } from "./config/env";
 import type { HarnessChatAgent } from "./features/ai-chat/harness-chat-agent";
+import {
+	type HarnessResumeStateStore,
+	PostgresHarnessResumeStateStore,
+} from "./features/ai-chat/harness-resume-state-store";
 import type { ArtifactDownloadSigner } from "./features/artifacts/artifact-download-signer";
 import type { ArtifactMetadataStore } from "./features/artifacts/artifact-metadata-store";
 import { PostgresArtifactMetadataStore } from "./features/artifacts/postgres-artifact-metadata-store";
@@ -38,6 +42,8 @@ export interface AppDeps {
 	config: ApiConfig;
 	/** Harness-hosted AI SDK chat turn (local composition only). */
 	harnessChatAgent: HarnessChatAgent;
+	/** Per-Conversation opaque Harness resume pointer (local composition only). */
+	harnessResumeStateStore: HarnessResumeStateStore;
 	/** Authoritative current Downloadable artifact metadata in Postgres. */
 	artifactMetadataStore: ArtifactMetadataStore;
 	/** Creates short-lived direct-download URLs after ownership authorization. */
@@ -100,6 +106,7 @@ export function createDeps(
 	return {
 		config,
 		harnessChatAgent,
+		harnessResumeStateStore: new PostgresHarnessResumeStateStore(database),
 		artifactMetadataStore: new PostgresArtifactMetadataStore(database),
 		artifactDownloadSigner,
 		conversationStore,
--- a/apps/chat-api/src/features/ai-chat/ai-chat.route.test.ts
+++ b/apps/chat-api/src/features/ai-chat/ai-chat.route.test.ts
@@ -26,16 +26,38 @@ const requestBody = {
 const uiMessageStream =
 	'data: {"type":"text-delta","id":"t1","delta":"Hi"}\n\ndata: [DONE]\n\n';
 
+/** What the fake session's `stop()` returns; opaque to the route. */
+const stoppedState = { type: "resume-session", data: { after: "turn" } };
+
+/** In-memory resume pointer store, optionally pre-seeded for conversation-1. */
+function fakeResumeStore(initial: unknown = null) {
+	const saved: unknown[] = [];
+	const store = {
+		load: async () => initial,
+		save: async (ref: unknown, state: unknown) => {
+			saved.push({ ref, state });
+		},
+	};
+	return {
+		store: store as unknown as AppDeps["harnessResumeStateStore"],
+		saved,
+	};
+}
+
 /** Fake agent that records the lifecycle and streams a canned UI message stream. */
 function fakeAgent() {
 	const events: unknown[] = [];
 	const created = {
-		destroy: async () => {
-			events.push("destroy");
+		stop: async () => {
+			events.push("stop");
+			return stoppedState;
 		},
 	};
 	const fake = {
-		createSession: async (options: { sessionId: string }) => {
+		createSession: async (options: {
+			sessionId: string;
+			resumeFrom?: unknown;
+		}) => {
 			events.push({ createSession: options });
 			return created;
 		},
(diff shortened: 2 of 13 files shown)


chengchao:
## Review of 2e40e18 (feat/harness-conversation-continuity vs origin/main)

Two-axis review (Standards / Spec) of the three commits since `origin/main`, against #599, parent #595, ADR-0033, and `docs/agents/*.md`.

Verified locally: `apps/chat-api` bun test 146 pass; `packages/agent-db` bun test 241 pass (13 skip, Postgres lane); `packages/agent-worker` cleanup tests pass; `biome check .` clean; `tsc --noEmit` in chat-api reports only the pre-existing unrelated `s3-artifact-download-signer.ts` S3Client error. The `@ai-sdk/harness` `createSession({ sessionId, resumeFrom })` / `stop(): Promise<HarnessAgentResumeSessionState>` contract matches what the route and store use.

### Spec (#599)

All eight acceptance criteria are met or deliberately deviated with a stated reason:

- Migration `0028_medical_jubilee` adds nullable jsonb `harness_resume_state`, owned by `packages/agent-db`, covered by `runtime-store.test.ts`.
- Route calls `stop()` on drained / cancelled / aborted / failed-to-start / mid-stream-failure paths (all six route tests updated) and stores the returned state.
- Stored pointer → `createSession({ sessionId, resumeFrom })`; none → fresh session, same id.
- Resume throw → fresh session, structured `warn`, 200 with a response (fake-agent test).
- Pointer is opaque and never enters the stream (asserted).
- No retention hook or sweep; docs state Vercel default.
- Local two-turn check documented in `docs/agents/chat-api.md`; port (3000 via compose), identity headers (`x-member-code`/`x-partner-code`), and the `POST /v1/conversations` body all check out against the code.

**Deviation ("deletes its row"):** the reasoning holds. `packages/agent-worker/src/cleanup/cleanup.ts:243-292` finds a deleted Conversation's E2B sandbox only through the surviving `conversation_runtime` row (left-join on missing `conversations`), kills it, then deletes the row itself. Dropping the row in `PostgresConversationStore.delete` would silently orphan any E2B sandbox and break the existing "leaves external cleanup state recoverable" test. Nulling `harness_resume_state` in the same transaction gives the pointer the exact lifetime the criterion wanted; the row goes with the next sweep. Approve the deviation; the PR body records it.

### Standards

No documented-standard violations. Drizzle is imported from the same instance as the rest of chat-api, the schema stays in `packages/agent-db`, the store is a thin unfenced upsert with the fencing rationale written down, tests use the PGlite harness, formatting/lint clean.

### Findings (none blocking)

1. **Low — `apps/chat-api/src/features/ai-chat/ai-chat.route.ts:110-121`.** The fallback retries `createSession` on *any* throw when a pointer exists, so a non-resume failure (Vercel auth/outage, `HarnessSandboxAuthenticationError`) is attempted twice and emits a misleading `harness session resume failed` warn before the real 500. This is literally what #599 asks for ("when resuming throws"), and the SDK does not expose a distinguishable "snapshot gone" error today, so accept as-is; note it in a follow-up if noisy in logs.
2. **Low / note — `postgres-conversation-store.ts:203-212`.** The AI SDK path has no Run, so `hasActiveRun` cannot see an in-flight Harness turn; a Permanent delete during a turn nulls the pointer and the turn's `stop()` then upserts it back onto the surviving runtime row. Harmless (the cleanup sweep removes the row; the sandbox expires per Vercel), and overlapping-turn/admission is an explicit follow-up slice of #595. No change needed now.
3. **Nit — `packages/agent-db/drizzle/meta/_journal.json`, `0028_medical_jubilee.sql`.** drizzle-kit output dropped the trailing newline on `_journal.json`. Cosmetic.
4. **Nit — `ai-chat.route.test.ts:32-46`.** `fakeResumeStore` returns `store as unknown as AppDeps["harnessResumeStateStore"]`; typing `load`/`save` against `HarnessResumeStateStore` directly would drop the double cast. Cosmetic.

Summary: Spec 0 gaps (1 approved deviation); Standards 0 violations, 4 non-blocking notes.

APPROVED FOR MERGE

Issue #635, opened 2026-08-30
How is a Harness turn interrupted, and what does aborting leave behind?


chengchao:
Part of #631

## Question

How is a Harness turn interrupted, and what does aborting mid-turn leave behind?

## Context

`POST /v1/conversations/:id/runs/:runId/interrupt` retires with the Run path (ADR-0013 covers interrupting Runs without ending Conversation continuity). The chat path has no equivalent endpoint — its only cancellation today is the request's abort signal, i.e. the client disconnecting.

The plumbing already exists end to end: chat-api passes `c.req.raw.signal` into `agent.stream`, and the bridge passes `abortSignal: abortCtl.signal` straight into `claudeSdk.query()` (`src/bridge/index.ts`). So the question is not *whether* interruption is possible.

## What has to be established

1. **What a client-triggered stop should look like.** Is disconnecting the stream sufficient for the assistant-ui client, or does the product need an explicit endpoint? Note that with one persistent sandbox per Conversation and an in-process `activeTurns` guard, a separate stop request must reach the same process that holds the turn.
2. **What abort leaves behind.** `cleanupAfterStream` runs `session.stop()` on cancel, which snapshots the sandbox and saves the resume pointer. Does a mid-turn abort snapshot a coherent state — a partial assistant message in the transcript, a half-written file, an interrupted tool call — and does the next turn resume cleanly from it?
3. **What gets persisted.** Feeds T5: on abort, does the user's message survive, and is a partial assistant message written or discarded?
4. **Whether the Agent SDK distinguishes** an aborted turn from a failed one on the stream, so the client can render the difference.

## Answer shape

A recommendation on the interrupt surface (endpoint or disconnect-only) plus the observed post-abort state of the snapshot, transcript, and resume path.

chengchao:
## Answer: disconnect is sufficient — but interruption is broken today

Note: [`docs/research/harness-turn-interruption-2026.md`](https://github.com/X-GPT/mymemo-agent/blob/research/harness-turn-interruption/docs/research/harness-turn-interruption-2026.md) on branch `research/harness-turn-interruption` (commit `a46fd78`), 503 lines, every claim cited to file and line.

### 1. A stop endpoint is not needed — disconnect is the AI SDK's own mechanism

[...]

PR #643, opened 2026-08-30
fix(chat-api): never persist a resume pointer left by a suspended turn

### LINKED ISSUE #642: Aborting a chat turn while a host tool is in flight wedges the Conversation permanently
## Symptom

Stopping a chat response can wedge a Conversation **permanently**. Every subsequent `POST /api/chat` on that Conversation returns 500 with `harness turn failed to start` in the logs. There is no recovery path — the Conversation stays broken for the rest of its life.

It reproduces most reliably in exactly the situation a user is most likely to press Stop: while a document tool is running. MyMemo's host tools are Postgres round trips, so a slow `SearchDocuments` or `LoadDocuments` is the window.

Diagnosed while resolving [How is a Harness turn interrupted, and what does aborting leave behind?](https://github.com/X-GPT/mymemo-agent/issues/635); full evidence in [`docs/research/harness-turn-interruption-2026.md`](https://github.com/X-GPT/mymemo-agent/blob/research/harness-turn-interruption/docs/research/harness-turn-interruption-2026.md) on branch `research/harness-turn-interruption`.

## Root cause

`HarnessAgentSession.stop()` branches on turn state (`@ai-sdk/harness@1.0.91`, `src/agent/harness-agent-session.ts:435-462`). If the framework still considers the turn running, it takes `suspendCurrentTurn()` rather than `doStop()` and returns a resume pointer carrying **`continueFrom`** — bridge coordinates `{ port, token, lastSeenEventId, sandboxId }`. Its `finally` block then **stops the sandbox anyway**, because chat-api passes a sandbox *provider* and never a session, so `ownsSandboxLifecycle` is `true`. The suspend branch exists precisely so the bridge stays alive; it is killed one line later.

The turn is still non-idle at that moment because `run-prompt.ts` awaits every outstanding host tool execution before settling the failure that returns turn state to idle.

Then, in [`apps/chat-api/src/features/ai-chat/ai-chat.route.ts`](apps/chat-api/src/features/ai-chat/ai-chat.route.ts):

1. **Line 159-160** — `stop()` saves whatever `session.stop()` returned, verbatim. `harness-resume-state-store.ts` is deliberately opaque and never inspects the payload, so the poisoned pointer is persisted.
2. **Next turn, lines 134-151** — the pointer is loaded and passed as `resumeFrom`. `HarnessAgent.createSession` lifts the nested continuation (`effectiveContinueFrom = validatedContinueFrom ?? validatedResumeFrom?.continueFrom`, `harness-agent.ts:281`) and constructs the session with `turnState: 'suspended'` (`harness-agent.ts:458-467`).
3. **Line 172** — chat-api calls `agent.stream`, which routes to `promptTurn`, whose first act is `requirePromptableTurn()` — and that throws on any non-idle state (`harness-agent-session.ts:585-595`).

**The resume fallback does not catch this.** The `try/catch` at lines 141-150 wraps `createSession` only, and `createSession` *succeeds* — it happily builds a suspended session. The throw comes from `agent.stream` at line 172, lands in the catch at line 177, is logged as `harness turn failed to start`, and becomes a 500. Because the poisoned pointer is never overwritten, this repeats forever.

## Reproduced

Against the real `HarnessAgentSession` and `runPrompt`, faking only the adapter and sandbox sessions:

| Scenario | `stop()` took | pointer carries `continueFrom` |
| --- | --- | --- |
| abort, 1 ms gap, no host tool in flight | `doStop` | no |
| abort, 1 ms gap, **host tool in flight (120 ms)** | `doSuspendTurn` | **yes** |
| abort, no gap | `doSuspendTurn` | **yes** |

Note the second row is the realistic case: a client disconnect fires `req.signal` roughly 1 ms before the response body's `cancel()` (measured on Bun 1.3.0), so `cleanupAfterStream` runs `stop()` while a tool call is still outstanding.

## Fix

**Do not persist a resume pointer that carries `continueFrom` — keep the previous one.** One condition around the save at `ai-chat.route.ts:160`:

```ts
const state = await session.stop();
if (state.continueFrom === undefined) {
  await c.var.deps.harnessResumeStateStore.save(ref, state);
}
```

### Do not "clean up" the pointer instead — it corrupts silently

The tidier-looking fix is to strip `continueFrom` and save the rest. **It is worse than the bug.** `toResumeStateWithContinuation` (`@ai-sdk/harness@1.0.91`, `src/agent/harness-agent-session.ts:572-583`) builds the poisoned pointer as:

```ts
return {
  type: 'resume-session',
  harnessId: continueFrom.harnessId,
  specificationVersion: continueFrom.specificationVersion,
  data: continueFrom.data,      // copied FROM the continuation
  continueFrom,
};
```

`data` is copied out of the continuation, because `doStop()` never runs on this branch — so it holds the suspend payload (dead bridge coordinates), not a resume payload. Removing `continueFrom` keeps the same broken address and deletes the only marker that says so, trading a loud 500 for silent corruption.

### Why skipping the save loses nothing

A healthy resume pointer carries **no state at all**. From `@ai-sdk/harness-claude-code@1.0.94`, `src/claude-code-harness.ts:790-798`:

> A `doStop()` payload is structurally empty (`{}`): the framework derives the sandbox via `provider.resumeSession({ sessionId })`, and the Claude SDK's `{ continue: true }` flag rehydrates the thread from the workdir.

The sandbox is found by `sessionId` — the Conversation id — and the transcript is rehydrated from the working directory inside the snapshot. There is no snapshot reference in the pointer to go stale, so retaining the previous pointer discards nothing. If no previous pointer exists (the first turn was stopped), nothing is saved and the next message starts a fresh session; the Conversation still works.

The check is also sound rather than heuristic: bridge coordinates appear **only** on `doDetach()` and `doSuspendTurn()` payloads and are documented as "Absent on a `doStop()` payload" (`claude-code-harness.ts:779-782`).

### On the resume store's opacity

`continueFrom` is an optional **top-level** field on `HarnessV1ResumeSessionState` (`src/v1/harness-v1-lifecycle-state.ts:43-51`), beside `harnessId`, `specificationVersion`, and `data` — not inside the adapter-defined `data` blob, which stays untouched. So reading it in the route barely touches the payload's opacity. Keep it simple; a `carriesUnfinishedTurn(state)` helper is likely over-engineering. Mention the choice in the PR either way.

## Regression test

`ai-chat.route.test.ts` already drives the route through a fake harness factory. Add a case where the fake session's `stop()` resolves to a pointer carrying `continueFrom`, and assert that:

1. the store is not asked to save that pointer (or saves the prior value), and
2. a second `POST /api/chat` on the same Conversation succeeds rather than 500ing.

Point (2) is the one that actually encodes the bug — assert it even though (1) is the mechanism.

## Scope

- Affects code shipped in #623 / #628 / #629. Reachable on the **local** composition today; `POST /api/chat` is not mounted in production, so no production impact yet.
- Independent of the production-readiness map ([#631](https://github.com/X-GPT/mymemo-agent/issues/631)) — it should not wait for it. The map records it in fog only as a pointer to this issue.
- Out of scope here: adopting the Agent SDK's `interrupt()` (available, since the bridge runs streaming-input mode, but unused) and any stop endpoint. #635 settled that disconnect is the right mechanism and no endpoint is needed.

chengchao:
Closes #642.

## What was broken

Stopping a chat response while a document tool was in flight wedged the Conversation permanently — every later `POST /api/chat` returned 500 with `harness turn failed to start`, with no recovery path.

`HarnessAgentSession.stop()` branches on turn state. When the framework still considers the turn running — which is exactly the case when a host tool is outstanding, the realistic Stop window — it takes `suspendCurrentTurn()` and returns a pointer carrying `continueFrom` (bridge coordinates), then stops the sandbox anyway in its `finally`, because chat-api passes a sandbox provider and owns the lifecycle. The route persisted that pointer verbatim; the next turn resumed into a `turnState: 'suspended'` session, and `agent.stream` → `requirePromptableTurn()` threw. The `createSession` resume fallback does not catch it — `createSession` succeeds; the throw comes later.

## The fix

Skip the save when the pointer carries `continueFrom`:

```ts
const state = await session.stop();
if (state.continueFrom === undefined) {
  await c.var.deps.harnessResumeStateStore.save(ref, state);
}
```

**Not stripping `continueFrom` and saving the rest.** `toResumeStateWithContinuation` copies `data` *out of the continuation*, so `data` is the suspend payload — the same dead bridge coordinates. Removing the field keeps the broken address and deletes the only marker that says so, trading a loud 500 for silent corruption.

**Keeping the previous pointer loses nothing.** A healthy `doStop()` payload is structurally empty: the sandbox is found by `sessionId` (the Conversation id) and the transcript is rehydrated from the workdir in the snapshot. There is no snapshot reference inside the pointer to go stale. If no previous pointer exists, the next message starts a fresh session and the Conversation still works. The check is sound rather than heuristic — bridge coordinates appear only on `doDetach()` / `doSuspendTurn()` payloads.

**On the store's opacity (per #642):** I took the plain inline check, no `carriesUnfinishedTurn(state)` helper. `continueFrom` is an optional *top-level* field on `HarnessV1ResumeSessionState`, beside `harnessId`/`specificationVersion`/`data` — the adapter-defined `data` blob stays untouched, so the payload's opacity is essentially intact and a one-implementation predicate would be over-engineering.

## Test

New case in `ai-chat.route.test.ts`. The fake session's `stop()` returns a pointer carrying `continueFrom`, and the fake agent mirrors the framework: `createSession` from such a pointer *succeeds*, and `stream` then throws. The load-bearing assertion is that a **second** `POST /api/chat` on the same Conversation returns 200 — verified to fail with 500 against the unfixed route. It also asserts the poisoned pointer never reaches the store; that a healthy pointer still does is already covered by the file's first test.

`fakeResumeStore` now reads back its last save; without that the regression test would have been tautological (the second turn would never load the pointer the first one wrote).

## Scope

Fix and test only. No `interrupt()`, no stop endpoint — #635 settled that client disconnect is the mechanism. `POST /api/chat` is not mounted in production, so no production impact.

Verified: `bun test` in `apps/chat-api` — 161 pass, 0 fail; `biome check` clean; `tsc --noEmit` clean for `src/features/ai-chat` (the one remaining error is the pre-existing dual-`@smithy/types` one in `s3-artifact-download-signer.ts`).

### DIFF
--- a/apps/chat-api/src/features/ai-chat/ai-chat.route.test.ts
+++ b/apps/chat-api/src/features/ai-chat/ai-chat.route.test.ts
@@ -377,28 +377,17 @@ it("keeps the previous resume pointer when a stopped turn was still running, so 
 		data: { bridge: "dead" },
 		continueFrom: { type: "continue-turn", data: { bridge: "dead" } },
 	};
-	let suspends = true;
 	let resumedSuspended = false;
 	fake.createSession = async ({ resumeFrom }) => {
 		// Mirrors the framework: creating a session from a pointer that carries
 		// `continueFrom` succeeds and yields a *suspended* session.
-		resumedSuspended =
-			(resumeFrom as { continueFrom?: unknown } | undefined)?.continueFrom !==
-			undefined;
-		return {
-			stop: async () => {
-				const state = suspends ? suspendedState : stoppedState;
-				suspends = false;
-				return state;
-			},
-		};
+		resumedSuspended = resumeFrom === suspendedState;
+		return { stop: async () => suspendedState };
 	};
 	fake.stream = async () => {
 		// ...and a suspended session refuses the next prompt, which is the 500.
 		if (resumedSuspended) throw new Error("turn is not promptable");
-		return {
-			toUIMessageStreamResponse: () => streamResponse(uiMessageStream),
-		};
+		return { toUIMessageStreamResponse: () => streamResponse(uiMessageStream) };
 	};
 	const { store, saved } = fakeResumeStore();
 	const app = makeApp({
@@ -418,14 +407,12 @@ it("keeps the previous resume pointer when a stopped turn was still running, so 
 	const first = await post();
 	expect(first.status).toBe(200);
 	await first.body?.cancel();
-	expect(saved).toEqual([]);
 
 	// Second turn on the same Conversation: answers instead of 500ing.
 	const second = await post();
 	expect(second.status).toBe(200);
 	expect(await second.text()).toBe(uiMessageStream);
-	// A pointer from a turn that did stop is still persisted.
-	expect(saved.map((s) => s.state)).toEqual([stoppedState]);
+	expect(saved).toEqual([]);
 });
 
 it("stops the session when the client cancels the stream", async () => {
--- a/apps/chat-api/src/features/ai-chat/ai-chat.route.test.ts
+++ b/apps/chat-api/src/features/ai-chat/ai-chat.route.test.ts
@@ -30,13 +30,17 @@ const uiMessageStream =
 	'data: {"type":"text-delta","id":"t1","delta":"Hi"}\n\ndata: [DONE]\n\n';
 
 /** What the fake session's `stop()` returns; opaque to the route. */
-const stoppedState = { type: "resume-session", data: { after: "turn" } };
+const stoppedState: { type: string; data: unknown; continueFrom?: unknown } = {
+	type: "resume-session",
+	data: { after: "turn" },
+};
 
 /** In-memory resume pointer store, optionally pre-seeded for conversation-1. */
 function fakeResumeStore(initial: unknown = null) {
 	const saved: { ref: unknown; state: unknown }[] = [];
 	const store = {
-		load: async () => initial,
+		// Reads back the last save, so a later turn sees what an earlier one left.
+		load: async () => saved.at(-1)?.state ?? initial,
 		save: async (ref: unknown, state: unknown) => {
 			saved.push({ ref, state });
 		},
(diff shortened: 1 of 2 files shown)


chengchao:
Closing: targets the abandoned Harness chat path — superseded by map #644.

PR #685, opened 2026-09-01
feat(in-vm-server): the trusted In-VM server serves one Turn (#662)

chengchao:
Closes #662. Parent: spec #654 (/v2 chat on Lambda MicroVMs).

## What this builds

The trusted **In-VM server** as a new workspace app (`apps/in-vm-server`), runnable locally against Postgres + Redis (no MicroVM yet — #666 bakes it into the image), serving one complete Turn:

- **Nudge (idempotent) + health** — the only endpoints. Idempotency lives in the DB's one-in-flight claim gate (`claimNextTurnTx`, #656); any number of nudges are safe, and a nudge while a Turn is processing (or with an empty queue) is a no-op.
- **One real SDK `query()` under the confinement bundle** (`src/query-options.ts`, pinned `@anthropic-ai/claude-agent-sdk@0.3.251` per the spec): `settingSources: []`, `strictMcpConfig`, `permissionMode: "dontAsk"`, cwd-scoped `Read(./**)`/`Edit(./**)` allows (never bare), OS-sandboxed Bash with network deny-all (`allowedDomains: []` + `strictAllowlist`) failing closed (`failIfUnavailable`, `allowUnsandboxedCommands: false`). The spawned CLI env is built from an allowlist — **no data-plane credential (agent DB / KB / Redis) can reach it**; the model credential rides along because in production it is the per-Conversation gateway token (env-configured `MODEL_BASE_URL`/`MODEL_API_KEY`/`MODEL`, no design change).
- **One assistant UIMessage per Turn** (`src/turn-stream-mapper.ts`): the SDK stream maps to stock AI SDK v7 `UIMessageChunk`s — `start`/`start-step`/`finish-step`, text/reasoning deltas, `tool-input-start` + `tool-input-available` (input arrives complete; skipping `tool-input-delta` is protocol-legal), `tool-output-available`/`tool-output-error`, terminal `finish` (Outcome in `messageMetadata`) / `error`. No custom chunk types; chunks validate at publish through the #658 relay lane. Durable parts come from the completed content blocks, never the deltas; subagent-internal traffic is filtered out.
[...]

Issue #668, opened 2026-08-31
v2: Interrupt a Turn


chengchao:
## Parent

Spec #654 — /v2 chat on Lambda MicroVMs.

## What to build

`POST /v2/conversations/:id/interrupt` — stop the currently processing Turn. 204 when nothing is processing (no turn id in the route — at most one Turn processes). Otherwise chat-api resolves the processing Turn's message id and sends the nudge's one command, `{interrupt: messageId}`; the In-VM server applies the SDK's `interrupt()` to the processing Turn — or, if the targeted Turn is still queued when the command arrives (race), cancels it directly. `interrupted` wins the Turn's Outcome once accepted. Interruption intent is ephemeral: no control table, no flag — a lost command is simply retried by the user; queued successors are never flushed.

## Acceptance criteria

- [x] Interrupting a long-running local Turn ends its SSE with `interrupted`, and the row terminalizes `interrupted` retaining exactly what durably completed.
- [x] Interrupt with nothing processing → 204, no side effects.
- [x] The race where the target is still queued: it terminalizes `interrupted` without ever running.
- [x] After an interruption, the next queued Turn serves normally.
- [x] No durable interrupt state anywhere; a dropped command leaves the Turn running and a retry succeeds.

## Blocked by

- #664 — In-VM server drain loop
- #667 — message POST + own-Turn SSE

chengchao:
## Verification — all five acceptance criteria met (PR #697, head `7f4ab57`)

| AC | Evidence |
|---|---|
| Interrupting a long-running Turn ends its SSE with the `abort` terminal chunk (`interrupted` Outcome) and the row terminalizes `interrupted` retaining exactly what durably completed | `apps/in-vm-server/src/turn-serving.test.ts` "terminalizes interrupted with exactly the committed Steps…": the committed Step is kept, the truncated envelope and a whole post-interrupt Step are neither persisted nor published, `abort` is the last chunk, and the stream drains to its result. Live probe against the **real Claude Code CLI** (real drain loop + `createAgentSession` + SDK `query()`, PGlite + in-memory relay, fake slow model): `turn-1 status: interrupted`, chunk tail `…text-delta,abort`, SDK tail after `interrupt()` = `assistant(aborted) → user → result(error_during_execution)` in 28 ms. |
| Interrupt with nothing processing → 204, no side effects | `conversation-messages.route.test.ts` "answers 204 with no side effects when nothing is processing": no nudge, rows unchanged. |
[...]

Issue #667, opened 2026-08-31
v2: message POST + own-Turn SSE


chengchao:
## Parent

Spec #654 — /v2 chat on Lambda MicroVMs.

## What to build

`POST /v2/conversations/:id/messages` — submit a UIMessage; the response is the standard UIMessage SSE scoped to the submitted message's own Turn: the **stock AI SDK v7 UI Message Stream** (https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol) — this route owns the SSE envelope (`x-vercel-ai-ui-message-stream: v1` header, `data:` framing, `[DONE]` terminator, keepalive pings) around the chunk payloads relayed from the #658 lane. Persist as `queued` → subscribe to the Turn's Live Stream BEFORE nudging → nudge → relay: silent SSE keepalives while queued predecessors drain, then only this Turn's events, ending at its terminal with the Outcome. There is no 409 anywhere — a second POST is simply the next queued row. A re-POST of the same client message id creates no second Turn. Client disconnect never interrupts the Turn (Recovery is from durable history at turn end). Guards carried from the v1 message path: internal identity, 404, archived-blocks-new-messages, exposure gate on submission. Ensure-VM is a dev-mode stub (config-provided In-VM server endpoint); the orchestration ticket replaces it — verified locally against the In-VM server from the drain-loop ticket.

## Acceptance criteria

- [x] Against a locally running In-VM server: POST streams the full Turn as UIMessage SSE, ends with the Outcome, and the durable history matches the streamed message.
- [x] The SSE is consumable by a stock v7 `useChat`/`DefaultChatTransport` with no custom transport — the Turn renders as one assistant message; asserted with the real client stream parser (`ai@7.0.83`) in a test.
- [x] Two rapid POSTs: no 409; the second holds with silent keepalives during its predecessor, then streams only its own Turn's events.
- [x] Subscribe-before-nudge ordering tested — no lost early events.
- [x] Duplicate client message id: no second Turn; the duplicate response's behaviour is documented and tested.
- [x] Client disconnect mid-stream leaves the Turn running to its Outcome (verified via durable history).
- [x] Archived Conversation refuses the message; the exposure gate is enforced on submission.

## Blocked by

- #657 — Conversation CRUD on /v2
- #658 — UIMessage Live Stream lane on the relay
- #664 — In-VM server drain loop

chengchao:
## Verification — all 7 ACs met (PR #694, head `32999e4`)

[...]

Issue #666, opened 2026-08-31
v2: In-VM server inside a real VM


chengchao:
## Parent

Spec #654 — /v2 chat on Lambda MicroVMs.

## What to build

Swap the image's placeholder for the real thing: bundle the In-VM server into the image as its entrypoint, wire platform-authenticated inbound (JWE) to nudge/health, deliver boot configuration (Conversation identity + the Gateway token) via `runHookPayload`, and serve a real Turn from inside a hand-launched VM in staging — the model call leaving via the Gateway route. Data plane reaches staging RDS/Redis as the topology allows (open managed egress is acceptable until the connector applies; the lockdown is verified at the pre-cutover gate). This ticket takes the first-contact risk of "real server on real platform", so orchestration doesn't have to.

## Acceptance criteria

- [ ] The CI image bakes the In-VM server; a hand-launched VM reaches ready with the server as its entrypoint.
- [ ] A nudge through the authenticated per-VM endpoint drains a queued Turn end-to-end from inside the VM — Live Stream and durable history land.
- [ ] The model call goes through the Gateway using the delivered token; the CLI env inside the VM carries no data-plane credential (asserted in staging).
- [ ] Sandbox-mode Bash engages in-VM (bubblewrap active, `failIfUnavailable` holds).
- [x] The boot sweep runs on server start.

## Blocked by

- #659 — Gateway route
- #661 — MicroVM image skeleton + authenticated boot
- #662 — In-VM server serves one Turn

chengchao:
## Status: PR ready, live ACs pending the runbook

PR #687 (head `dee50c0`, one squashed commit on main) is merge-ready: Ponytail clean over 7 review rounds, an APPROVED FOR MERGE code-review verdict, and CI green including the image contract check. Not merged yet — the live ACs below gate that call.

### AC-by-AC evidence

- [ ] **CI image bakes the In-VM server; a hand-launched VM reaches ready with the server as its entrypoint** — *half proven.* The bake half is done: the PR's `image-contract` check builds the staged ARM64 context (pruned workspace, `bun install --production --filter in-vm-server`, bun 1.3.0) and boots the baked entrypoint unconfigured, probing `/ready` 200, `/health` 200, `/nudge` 503. The hand-launched-VM half needs the registered image (CI registers on the merge push).
[...]

PR #686, opened 2026-09-01
feat(in-vm-server): drain loop with one long-lived Agent session (#664)

### LINKED ISSUE #664: v2: In-VM server drain loop
## Parent

## What to build

The In-VM server becomes the Conversation's serializer: feeder/consumer tasks over the DB queue, one-in-flight gated by the DB itself (push next only when nothing is processing), the feeder on a wakeable sleep (nudge doorbell + interval self-heal), ONE long-lived `query()` carrying the Agent session across Turns, and an `error` Outcome that never wedges the loop.

## Acceptance criteria

- [x] Three queued Turns and a single nudge → all three serve strictly in submission order, one at a time.
- [x] A Turn ending `error` leaves the loop healthy — the next queued Turn serves.
- [x] No lost work: a message enqueued mid-Turn is picked up without another nudge (interval self-heal), and a nudge arriving mid-Turn wakes the feeder after the terminal.
- [x] Model-side memory persists across Turns within one process lifetime (a later Turn sees an earlier Turn's context through the long-lived session).
- [x] Restarting the server with queued rows resumes draining (boot sweep + claim on start).

## Blocked by

chengchao:
Closes #664. Parent: spec #654 (/v2 chat on Lambda MicroVMs). Builds on #662 (one-Turn serving, PR #685).

## What this builds

The In-VM server becomes the **Conversation's serializer** — two new modules, `serveOneTurn` untouched:

- **`src/drain-loop.ts` — the feeder.** One task claims and serves queued Turns strictly in sequence order; one-in-flight stays gated by the DB itself (`claimNextTurnTx` refuses while anything is `processing`, #656). With the queue empty it parks on a **wakeable sleep**: a nudge rings the doorbell for an immediate re-check (a nudge arriving mid-Turn is kept and consulted right after the terminal — the doorbell check and the wake installation share one synchronous executor, so no nudge can fall between them), and the **interval tick self-heals a lost nudge**. A Turn's `error` Outcome never wedges the loop — the next claim proceeds regardless; a thrown `serveOneTurn` (a claim-time DB blip) is logged and retried on the sleep cadence. On start the loop **boot-sweeps** stale `processing` rows to `interrupted` (retried until it lands — claiming is gated on nothing `processing`, so draining can't start ahead of it) and then claims immediately, so a restart with queued rows resumes draining without a nudge.
- **`src/agent-session.ts` — ONE long-lived `query()`.** Streaming-input mode keeps the CLI process and its model-side context alive between Turns. `createAgentSession` adapts that single stream to `serveOneTurn`'s per-Turn `TurnQueryFn`: each call pushes one user message into the shared prompt channel and yields the session's messages up to and including that Turn's `result` (windows never overlap — the DB gate serializes callers). A session that breaks — the stream throws, ends result-less, or a window is abandoned before its `result` (a protocol violation or persistence failure inside `serveOneTurn`) — is **retired, closing the underlying stream**, and the next Turn lazily starts a fresh `query()`; the shared stream is never left mid-Turn for the next window to read the previous Turn's tail. Model-side memory resets with it: the guarantee is process-lifetime and best-effort, never durable.
- **Wiring** (`src/index.ts`): the nudge endpoint now rings the drain loop's doorbell instead of firing `serveOneTurn` directly.

## Acceptance criteria → tests

- **Three queued Turns, a single nudge → all three serve strictly in submission order, one at a time**: `drain-loop.test.ts` parks the loop (interval effectively disabled), enqueues three, nudges once, and asserts served order and all-`done` (the DB gate makes serving serial).
- **A Turn ending `error` leaves the loop healthy**: an `error_during_execution` result on Turn 1 still drains Turn 2 to `done`.
- **No lost work**: (a) a Turn enqueued mid-Turn (from inside the serving fake, its nudge rung mid-Turn) serves right after the terminal with no further nudge; (b) a Turn enqueued while parked with the nudge lost entirely serves on the interval tick.
- **Model-side memory persists across Turns**: the honest composition — drain loop over `createAgentSession` over a context-carrying fake session stream — shows Turn 2's durable assistant text containing Turn 1's prompt, with exactly ONE underlying `query()` call. `agent-session.test.ts` additionally pins: consecutive windows off one query, prompts in order on one iterable; a thrown stream fails only its window and the next Turn gets a fresh query; a result-less stream end retires the session; an abandoned window closes the underlying stream before the next window starts.
- **Restart resumes draining**: a seeded stale `processing` row plus two queued rows, loop started with no nudge → stale row `interrupted` (never re-served — at most once), queued rows drain in order.

## Notes for review

- The prompt channel buffers at most one message (the DB admits one Turn at a time); buffering exists so a push is safe even before the SDK first pulls.
- `SDKUserMessage` needs only `type`/`message`/`parent_tool_use_id` (`uuid`/`session_id` optional in the pinned `0.3.251` typings).
- The session ignores the per-call `options` `serveOneTurn` passes — they are the same object as the bundle fixed at session start; noted at the seam.
- Self-heal interval is a constant (15 s) with a deps override for tests — no new env/config surface.
- Interrupt (nudge body command) is #668; JWE inbound + image bundling are #666; graceful drain on `/suspend` is #667. Shutdown behavior is unchanged from #662.

## Verification

- `apps/in-vm-server`: 64 pass (54 before); `bunx tsc --noEmit` clean; `bunx biome check src/` clean.
- Full root suite (`bun run test`): all workspaces green (agent-db 254 pass / 17 postgres-lane skips, agent-worker 30, agentcore-dispatch 9, live-text 51).

https://claude.ai/code/session_01C7JE2DPgHkiQ5mAkDmpARW


chengchao:
## Code review (head ff6af82)

Two-axis review of `main...HEAD` (500aaf3, ff6af82): **Standards** (docs/agents/*.md + smell baseline) and **Spec** (issue #664 acceptance criteria). Each finding below was re-verified against the code; findings that failed verification were dropped and are noted. Verification: `bun test` in `apps/in-vm-server` (64 pass), `bunx tsc --noEmit` clean, `bunx biome check src/` clean.

### Standards

No hard violations of a documented standard. The architecture module map is updated, the turn-store boundary is respected, domain vocabulary matches the glossary, the confinement bundle stays a single fixed `queryOptions`, and the Drizzle/pglite peer invariant (docs/agents/development.md) already holds in `apps/in-vm-server/package.json`.

Judgement-call notes (non-blocking, no change required):

- **Duplicated test harness** — `drain-loop.test.ts:52-104` re-declares `turn-serving.test.ts`'s `silentLogger`/`SENTINEL_OPTIONS`/`enqueue`/`makeDeps`/DB-seed shapes near-verbatim (`SENTINEL_OPTIONS` now lives in three files). `src/testing/` is the established home if extracted — but working-agreements.md's "don't refactor adjacent code" argues for leaving it this ticket. Note for whoever touches these tests next.
- **Ignored per-call `options`** — `agent-session.ts:41-46`: `createAgentSession` implements `TurnQueryFn` but ignores the per-call `options` (documented in a comment; `index.ts:32-33` wires the identical object). Harmless today; a future caller passing different options gets no signal. A dev-time assert or a narrower session-facing type would close it.

Dropped after verification (would not be acted on): wake-idiom duplication across `agent-session.ts`/`drain-loop.ts` (semantics differ — extraction would be speculative), the pre-existing `{userId, conversationId}` clump, `Awaited<ReturnType<typeof serveOneTurn>>` spelling, and the repeated `Bun.sleep(20)` test warm-up.

### Spec (issue #664)

All five acceptance criteria are implemented and covered by real DB-backed tests in `drain-loop.test.ts`: ordered drain ("serves three queued Turns strictly in submission order"), error never wedges ("serves the next queued Turn after a Turn ends error"), no lost work (mid-Turn enqueue + "the interval tick self-heals a lost nudge"), cross-Turn memory (drain loop composed over the real `createAgentSession`, `sessionCalls === 1`, second assistant row carries the first Turn's context), and restart resume ("boot-sweeps the stale processing Turn and drains queued rows with no nudge", including the never-re-run assert).

Non-blocking findings:

- **Info — feeder/consumer collapsed into one task** (`drain-loop.ts:78-107`). The spec says "feeder/consumer tasks over the DB queue"; the implementation is one sequential task that claims and serves inline. Behaviorally equivalent — "one-in-flight gated by the DB itself" makes a separate consumer pure overhead, and every AC holds — but it's a visible shape deviation worth one sign-off line on the ticket.
- **Minor — `shutdown()` never calls `loop.stop()`** (`index.ts:42-48` vs `drain-loop.ts:110-115`). SIGTERM mid-Turn exits without draining; `stop()` is only exercised by tests. Recovery is exactly AC5's boot sweep (stale `processing` → `interrupted` on restart), and waiting out an in-flight Turn in a signal handler is its own can of worms, so this is acceptable as-is — but it's worth a deliberate decision rather than an accident.
- **Medium, pre-existing (follow-up ticket, not this PR)** — a failed `terminalizeTurnTx` on the error path (`turn-serving.ts:147-156`, #662 code) leaves the row `processing`; `claimNextTurnTx` then refuses every claim and `sweepStaleProcessingTurnsTx` only runs at boot, so the Conversation stays wedged for the rest of the process lifetime while the interval spins. The same double-failure wedged the old per-nudge wiring too — #664 actually adds the restart heal — but the loop makes the wedge permanent-until-restart and nothing retries the terminalization. Suggest a follow-up issue.
- **Minor — the mid-Turn doorbell branch isn't distinguishably tested.** `drain-loop.ts:103`'s `|| doorbell` is masked by `outcome !== null` (every served Turn already forces a re-claim); the mid-Turn test (`drain-loop.test.ts:179-202`) passes with the doorbell check deleted. The doorbell is load-bearing only in the `serveOneTurn`-threw path (`outcome === null`), which no test exercises. The code is correct (the doorbell-check/wake-install share one synchronous executor — verified race-free); this is a test-evidence gap, and the test's "without another nudge" title doesn't match its body (it does fire `loop?.nudge()` mid-Turn; the lost-nudge clause is the self-heal test's job).

Dropped after verification: a claimed CLI-process leak when the underlying stream throws (`agent-session.ts:97-101` calling `iterator.return?.()` on a completed generator). The real SDK `Query.return()` is a custom method — `async return(e){ return await this.cleanup(), this.sdkMessages.return(e) }` — that always runs cleanup (transport close + bounded wait for child exit) regardless of generator state, so the retire path does reap the CLI.

### Summary

Standards: 0 hard violations, 2 non-blocking notes (worst: test-harness duplication). Spec: 4 non-blocking findings (worst: the pre-existing terminalize double-failure wedge — follow-up ticket material). No verified finding requires a code change before merge.

APPROVED FOR MERGE

chengchao:
Response to the review (head unchanged at ff6af82 — all findings non-blocking, none require a code change):

- **`shutdown()` not calling `loop.stop()` is now a deliberate decision**: SIGTERM exits mid-Turn by design at this ticket; the durable recovery is exactly AC5's boot sweep, and the graceful-drain gate belongs to #667's `/suspend` hook (holds while `processing`, then checkpoints), which is the right place to wait out an in-flight Turn.
- **Feeder/consumer collapsed into one task**: signed off on the ticket — https://github.com/X-GPT/mymemo-agent/issues/664#issuecomment-5490295310.
- **Pre-existing terminalize double-failure wedge** (#662 code): agreed it's follow-up ticket material, not this PR — the failure needs a retry story around `terminalizeTurnTx`, orthogonal to the drain loop.
- **Doorbell-branch test evidence + mid-Turn test title**: acknowledged as a test-evidence gap, not a behavior bug (the branch is load-bearing only in the `serveOneTurn`-threw path); noted for the next touch of these tests along with the harness-duplication note.

Issue #670, opened 2026-08-31
v2: Checkpoint + lifecycle hooks


chengchao:
## Parent

Spec #654 — /v2 chat on Lambda MicroVMs.

## What to build

Durability across VM replacement. The suspend hook becomes a graceful-drain gate: it holds while a Turn is processing, then writes the FULL Checkpoint — the Agent session (`~/.claude`) and the Workspace — to the Conversation's S3 prefix. The suspend-time Checkpoint is the durable one and always complete: terminating a suspended VM fires no hook. Restore-on-boot completes before the VM reports ready; rehydration (a fresh VM restoring the Checkpoint) is how Conversations survive the 8-hour cap and how image upgrades land. The boot sweep is wired into the real lifecycle; the Checkpoint pointer is maintained on the VM record.

## Acceptance criteria

- [x] Suspend→resume preserves the Agent session and Workspace natively (platform snapshot; markers survive).
- [x] A suspend arriving mid-Turn holds until the Turn's terminal, then the Checkpoint completes to the Conversation's prefix.
- [x] Terminate the suspended VM → next message rehydrates: Workspace and Agent session restored before ready, and model-side memory continuity across the replacement is demonstrated in a real exchange.
- [x] A hard kill mid-Turn → the replacement VM's boot sweep terminalizes that Turn `interrupted`; history shows exactly what durably completed.
- [x] The Checkpoint pointer on the VM record tracks the latest durable Checkpoint.

## Blocked by

- #666 — In-VM server inside a real VM
- #669 — orchestration: claim, launch, rehydrate

chengchao:
Two findings from #666's live verification that change this ticket's approach before it's built.

## The VM cannot reach S3

There are no VPC endpoints anywhere in the agent Terraform stack, and `infra/terraform/microvm.tf`'s route tables carry only the implicit VPC-local route — that absence *is* the egress lockdown. The MicroVM security group permits exactly RDS ×2, Redis, ALB:80, and DNS. So the design as written here (the VM's execution role writing checkpoints to `conversations/*`) cannot work: there is no network path to S3, with or without credentials. Presigned URLs don't help either — they still need S3 reachability.

Adding an S3 gateway endpoint would work but punches a new data path through the lockdown this architecture is built around.

## Broker the checkpoint through chat-api instead

[...]

Issue #662, opened 2026-08-31
v2: In-VM server serves one Turn


chengchao:
## Parent

Spec #654 — /v2 chat on Lambda MicroVMs.

## What to build

The trusted In-VM server as a new workspace app, runnable locally against Postgres + Redis (no MicroVM yet), serving one complete Turn: on nudge it claims a single queued Turn, runs one real SDK `query()` turn under the confinement settings bundle (`settingSources: []`, `strictMcpConfig`, cwd-scoped Read/Edit, OS-sandboxed Bash with network deny-all, `failIfUnavailable`), maintains the Turn's **single assistant UIMessage row** in `conversation_messages` — upserted at each Step's completion boundary, always BEFORE that Step's completion chunk publishes on the v2 Live Stream lane — and terminalizes the Turn with its Outcome. Endpoints: nudge (idempotent) and health. Model access is an env-configured base URL/credential in the trusted process — locally direct or via the Gateway route; in production this becomes the per-Conversation gateway token, with no design change here. The spawned CLI's environment carries no data-plane (agent DB / KB / Redis) credential.

## Claude SDK → UIMessage chunk mapping (amended 2026-08-31 — one UIMessage per Turn)

The consumer maps Claude Agent SDK stream output to **stock AI SDK v7 UI Message Stream chunks** (https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol; validated with `ai@7.0.83`'s `UIMessageChunk` types, per #658). There is no `toUIMessageStreamResponse()` in-VM: that helper wraps an AI SDK core stream, which the Claude SDK is not — the mapping is hand-written here.

**Framing: the whole Turn is ONE assistant UIMessage.** Each provider response is a **Step** inside it (`start-step`/`finish-step`), tool parts embedded — the framing AI SDK's own `ToolLoopAgent`/`WorkflowAgent` use, and the only one stock `useChat` can render (source-verified: `processUIMessageStream` holds one assistant message per stream; a second `start` mutates its id, never opens a new message). This keeps stream shape ≡ history shape for stock clients.

**Durability follows the Step**: the Turn's single assistant row upserts with its completed parts at each `finish-step` boundary — commit-before-publish applies per Step. The Turn's terminal chunk publishes only after the final upsert and the status flip. An interrupted or error Turn retains exactly its completed Steps; a Step in flight when the Turn dies is never persisted (no provisional text).

| Claude SDK event | v7 chunk(s) |
| --- | --- |
[...]

Issue #661, opened 2026-08-31
v2: MicroVM image skeleton + authenticated boot


chengchao:
## Parent

Spec #654 — /v2 chat on Lambda MicroVMs.

## What to build

The production MicroVM image pipeline, proven with a placeholder server so the platform risk retires before the real In-VM server exists. Start from the probe kit on the `research/microvm-probe` branch (Dockerfile + minimal health/hooks server + RESULTS). Image: AL2023 base at DEFAULT capabilities (bubblewrap needs no `ALL` flag — proven live), nodejs22 + bubblewrap + socat + shadow-utils, pinned `@anthropic-ai/claude-agent-sdk@0.3.251` and CLI `2.1.251`, root-owned managed settings with `Edit(...)` deny rules (`Write(...)` entries are inert — omit them), non-root user. ARM64 image build in CI via `create-microvm-image`. Apply the probe's recorded platform corrections (full ARN for image lookup, hyphenated build log group, dnf's all-or-nothing installs, transient 502 on launch).

## Acceptance criteria

- [ ] CI builds and registers the ARM64 image on main.
- [ ] A hand-launched VM from the image reaches ready; health answers through the platform's JWE-authenticated per-VM endpoint.
- [ ] Managed settings baked root-owned and non-writable by the runtime user (spot-checked in-VM).
- [ ] Lifecycle hooks answer so boot/suspend/resume/terminate cycle cleanly.
- [ ] SDK/CLI versions pinned in the image; bubblewrap functional inside the VM (namespace smoke).

## Blocked by

None — can start immediately.

chengchao:
## Acceptance verified live — 2026-09-01 (us-west-2)

Shipped via #679 (image kit + CI pipeline + build role + deploy-role grants), plus two live IAM corrections the first register runs surfaced: #682 (`CreateMicrovmImage` authorizes against resource `*`) and #683 (`PassNetworkConnector` on the managed connector — image builds pass `INTERNET_EGRESS` implicitly). Both applied to the bootstrapped deploy role.

- [x] **CI builds and registers the ARM64 image on main** — [register run](https://github.com/X-GPT/mymemo-agent/actions/runs/33461619930) on merge commit `9e55a0c`: `MicroVM image mymemo-agent-prod-microvm is CREATED` (~3 min build).
[...]

PR #687, opened 2026-09-01
feat(in-vm-server): the In-VM server inside a real VM (#666)

### LINKED ISSUE #666: v2: In-VM server inside a real VM

chengchao:
Closes #666. Parent: spec #654 (/v2 chat on Lambda MicroVMs).

## What this builds

The MicroVM image's placeholder retires: the real **In-VM server** (`apps/in-vm-server`, #662) is now the image entrypoint, with platform-authenticated inbound and boot configuration via `runHookPayload`.

- **Boot model (one config contract, two delivery modes)**: the image snapshots the server *unconfigured* — it listens, answers the build's `/ready` hook, and 503s `/nudge`. At `run-microvm` the platform POSTs `/aws/lambda-microvms/runtime/v1/run` with the `runHookPayload` — a JSON object of exactly the env keys the server already reads locally (`AGENT_DATABASE_URL`, `REDIS_URL`, `MYMEMO_USER_ID`, `MYMEMO_CONVERSATION_ID`, `MODEL_*`, …) — and the server configures Turn serving, **runs the boot sweep** (stale `processing` → `interrupted`, never re-run), exec-verifies the CLI binary, and only then returns 200. The platform gates all endpoint traffic until `/run` returns, so a nudge can never reach an unconfigured VM; a malformed payload fails the hook (non-200) and the VM never serves. Locally, `MYMEMO_CONVERSATION_ID` in the plain environment selects immediate configuration at startup — the #662 dev flow is unchanged, and the boot sweep now runs there too.
- **HTTP surface** (`src/app.ts`): nudge + health as before, plus the platform lifecycle hooks (`run|ready|validate|resume|suspend|terminate`, POST, 200 = proceed; graceful-drain suspend + checkpointing are #670), and `GET /smoke` (image-only, `SMOKE_SCRIPT` env) streaming the baked in-VM acceptance checks. Everything arrives through the platform's JWE-authenticated per-VM endpoint (`X-aws-proxy-auth`) — the proxy is the authentication boundary.
- **CLI binary resolution** (`src/claude-code-executable.ts`): deliberate duplicate of agentcore-runtime's resolver (AgentCore retires wholesale at cutover, ADR-0034 — surviving code must not depend on it, same stance as the gateway route's non-shared stream helper). Resolves the SDK's **glibc** linux-arm64 optional-dependency binary (the SDK's musl-first default cannot execute on AL2023), exec-verifies it, and passes it as `pathToClaudeCodeExecutable` in the confinement bundle.
- **Image packaging** (`infra/microvm-image/Dockerfile` + `scripts/deploy/stage_microvm_image_context.sh`): pruned-workspace production install (`bun install --frozen-lockfile --production --filter in-vm-server`, bun pinned 1.3.0 = the CI pin), mirroring agentcore-runtime's proven pattern. One staging script assembles the identical context for the PR-time docker build and for registration, so what CI checks is byte-for-byte what registers. The global npm SDK/CLI installs are gone — one serving copy, attested by `smoke.sh` (SDK `0.3.251` / CLI binary `2.1.251` via the SDK manifest, ELF check for the arm64 binary). Adds the digest-pinned RDS CA bundle (`NODE_EXTRA_CA_CERTS`), a `developer`-writable Workspace, and a root-owned (non-writable) server install.
- **Pipeline**: `microvm-image.yml` triggers now include the server sources; the PR contract check **boots the baked server unconfigured** in the ARM64 container and probes `/ready` 200, `/health` 200, `/nudge` 503 — a broken install fails the PR, not the registration. `register_microvm_image.sh` stages + zips the context and takes `MICROVM_IMAGE_NAME` for scratch pre-merge builds.

> **Post-review updates**: #664 (drain loop) landed mid-review — `configure()` now composes the long-lived Agent session + drain loop, with one awaited boot sweep before `/run` answers 200 (an unreachable data plane fails the launch loudly; the loop's retried sweep then finds nothing). Ponytail review rounds trimmed the resolver copy to its one caller, dropped the dead `validate` hook route, and removed runtime-dead files from the image. The branch was then rebased onto main as one squashed commit — its tree is byte-identical to reviewed head `991085d` (both review verdicts and the green CI ran on that exact tree).

## Verification

- `apps/in-vm-server`: 80 pass (hooks, payload parsing, real-resolution resolver check, confinement bundle incl. `pathToClaudeCodeExecutable`, drain loop + agent session from #664); `bunx tsc --noEmit` clean; biome clean.
- Full root suite (`bun run test`): all workspaces green.
- **Offline image contract (run locally, ARM64 via qemu)**: full docker build of the staged context, then `scripts/smoke/microvm-image-check.sh` — every smoke RESULT PASS (pins on the serving install, ELF arm64 CLI binary, policy tier root-owned + non-writable, Workspace writable, app install non-writable) and the baked server **boots unconfigured and answers `/ready`/`/health`/`/nudge` exactly per the run-hook contract**.
- Production filtered install proven from the staged context (drizzle-orm resolves via agent-db's peer set; the SDK's platform binary resolves and exec-verifies through the isolated-linker store).

## Live verification (remaining, needs operator credentials)

This session's permission mode blocks prod-account writes (image registration, `ecs run-task` seeding, Secrets Manager reads), so the on-platform ACs are staged for the operator instead of verified here:

- `infra/microvm-image/README.md` carries the full hand-launch runbook: register (CI on merge, or `MICROVM_IMAGE_NAME=… register_microvm_image.sh` for a scratch build), launch through the `mymemo-agent-prod-microvm-egress` connector with the execution role and a real `runHookPayload`, then health/smoke/nudge through the JWE endpoint and a real Turn through the `/v2/gateway` route.
- Suggested end-to-end probe for the credential ACs: make the verification Turn run `env | sort` via sandboxed Bash — the durable history must show no data-plane credential in the CLI env, and the gateway's attribution log line proves the model call went through the delivered token.

## Notes for review

- `/smoke` sits beside the spec's nudge + health because #661 shipped it deliberately: content-free diagnostics behind the JWE proxy, and the only way to re-prove bwrap/policy-tier ACs on a live VM without shell ingress. It does not register locally.
- `PORT`/`LOG_LEVEL` moved out of `InVmConfig` to the entrypoint: the listener must exist before any Conversation is assigned (the build's `/ready` fires with no configuration at all), so they can never ride the payload.
- The RDS CA bundle is fetched digest-pinned by the staging script rather than `ADD --checksum` — the Lambda image builder's BuildKit frontend support is unverified (older local BuildKit already rejects the flag).

https://claude.ai/code/session_01FBtkhfv1sRxGbXF9G8xz37


chengchao:
Independent review of `feat/666-in-vm-server-in-real-vm` (head 6632471) against merge-base `56a400b` (origin/main), two axes: repo standards (CODING_STANDARDS.md + docs/agents/\*) and spec (#666, parent #654). Verification re-run locally: `bun test` in apps/in-vm-server 80 pass / 0 fail, `bunx tsc --noEmit` clean, biome clean.

The core implementation is sound: `envFromRunHookPayload` fails the `/run` hook loudly on anything malformed, an unconfigured `/nudge` answers 503, the CLI env allowlist keeps every data-plane credential out of the spawned CLI while the gateway token rides exactly as #654 prescribes, the pins are exact (SDK 0.3.251, CLI 2.1.251 via the SDK's linux-arm64 optional dep), the glibc-not-musl resolver is justified and exec-verified, and the CI contract check boots the actual baked entrypoint unconfigured — a broken install fails the PR, not the registration.

## Blocking (doc accuracy — both introduced by this diff, both small fixes)

1. **The runbook misstates the `/run` boot contract.** `infra/microvm-image/README.md` (line 7, "Boot model") and the PR body claim the server "configures Turn serving, runs the boot sweep, exec-verifies the CLI binary, and **only then returns 200**". That was true at commit 6e76e07 (sweep awaited before `serving` was assigned), but after the #664 drain loop merged in, `configure()` in `apps/in-vm-server/src/index.ts` calls `startDrainLoop()`, which returns synchronously — the sweep runs asynchronously inside the loop, retried until it lands, and `/run` returns 200 without waiting for it. The at-most-once invariant still holds (claiming is gated behind the sweep in `drain-loop.ts`), but two consequences need an explicit decision:
   - The README is the operator runbook the still-pending live ACs depend on, and it now asserts a false ordering.
   - Commit 6e76e07's stated property — "a failed configure must leave the server unconfigured so a retried /run reconfigures instead of answering 'already configured' over broken dependencies" — is silently gone: a well-formed payload with an unreachable/wrong `AGENT_DATABASE_URL` now yields a 200 run hook and a permanently "serving" VM that retries the sweep forever, instead of a failed launch.

   Fix either direction: correct the README/PR prose (sweep is started and gates claiming, not the 200) and accept the retry-forever behavior explicitly, or await the first sweep in the `/run` path before assigning `serving` so DB-level misconfiguration fails the hook loudly, as the prose promises.

2. **`docs/agents/configuration.md` baked-env list is incomplete.** The section this diff rewrote says "The image bakes only shared, non-secret environment (`WORKSPACE_DIR`, `NODE_EXTRA_CA_CERTS`, `CLAUDE_CODE_DISABLE_AUTO_MEMORY`)", but the Dockerfile also bakes `SMOKE_SCRIPT=/opt/microvm/smoke.sh`, which `src/index.ts` reads to register `GET /smoke`. Add it to the list.

## Non-blocking

- `smoke.sh`'s `cli-pinned` check now reads the version from the SDK's `manifest.json` instead of executing the CLI (the #661 check ran `claude --version`). The ELF check plus `/run`-time `execFileSync` cover executability, but in-VM (where qemu is no excuse) the smoke could exec the resolved binary for its version — the pin AC is currently attested by metadata.
- `GET /smoke` + `SMOKE_SCRIPT` is a standing deviation from #654's endpoint contract ("nudge … and health. Nothing else."). Deliberate since #661, content-free, behind the JWE proxy — fine, but worth a footnote on the spec/ADR so the contract doc matches reality.
- The resolver copy (`claude-code-executable.ts`) drops agentcore-runtime's try/catch around the exec-verify, so a present-but-broken binary surfaces as a raw `execFileSync` error rather than a labeled one. The `/run` hook still logs and fails; acceptable.
- `{ microvmId?: string; runHookPayload?: string }` is spelled twice in `app.ts` (the `AppHandlers.run` signature and the route's cast) — a named type keeps them from drifting.
- `rds_bundle_sha256` in `stage_microvm_image_context.sh` duplicates the pin in `apps/agentcore-runtime/Dockerfile`; a bundle rotation touches both. Transient until AgentCore retires at cutover.
- `query-options.test.ts`'s `pathToClaudeCodeExecutable` assertion is an input-echo (borderline under CODING_STANDARDS.md's tautological-tests rule); defensible as pinning that the option lands in the confinement bundle.

## Live-verification gap (documented, not a blocker)

Four of the five #666 ACs (hand-launched VM reaches ready, nudge drains a Turn end-to-end, credential-free CLI env asserted in staging, bwrap engaging in-VM) require prod-account access this session couldn't use; the PR says so and stages the operator runbook in `infra/microvm-image/README.md`. The offline substitutes (CI boot-contract check, smoke.sh assertions, unit tests) are real and correct. #666 should stay open past merge until the runbook is executed — which is exactly why blocking finding 1 (the runbook's accuracy) is worth fixing first.

Out-of-scope tickets (#664 drain loop, #665 doc tools, #670 suspend/checkpoint, #668/#669 orchestration) were treated as such; their absence is not a finding.

CHANGES REQUIRED

chengchao:
Independent re-review of `feat/666-in-vm-server-in-real-vm` (head 991085d) against merge-base `56a400b` (origin/main), two axes: repo standards (CODING_STANDARDS.md + docs/agents/\*) and spec (#666, parent #654). Verification re-run locally: `bun test apps/in-vm-server` 80 pass / 0 fail, biome clean.

## Prior blocking findings — both verified fixed on this head

1. **Awaited boot sweep before `/run` answers 200.** `configure()` in `apps/in-vm-server/src/index.ts` now runs `await sweepStaleProcessingTurnsTx(...)` and assigns `serving` only after it resolves; the `/run` handler awaits `configure()` and any throw becomes Hono's 500 (covered by the "failing run handler answers non-200" test), so an unreachable data plane fails the launch loudly and leaves the server unconfigured for a retried hook. Local mode top-level-awaits the same `configure()`, so the sweep runs in both delivery modes — the "boot sweep runs on server start" AC holds, and the README/PR boot-model prose is now accurate.
2. **`SMOKE_SCRIPT` documented.** `docs/agents/configuration.md` now lists it in the baked image env alongside `WORKSPACE_DIR`, `NODE_EXTRA_CA_CERTS`, `CLAUDE_CODE_DISABLE_AUTO_MEMORY`, matching the Dockerfile.

## Spec (#666 / #654)

Traced clean: malformed `runHookPayload` fails the hook non-200 (`envFromRunHookPayload` rejects non-string/bad-JSON/non-object/non-string-values; tested); `/nudge` answers 503 until `serving` is assigned; the gateway token rides `MODEL_BASE_URL`/`MODEL_API_KEY` into `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` with `ANTHROPIC_API_KEY` pinned empty and the six-key allowlist keeping DB/Redis credentials out of the CLI env; the confinement bundle matches #654 exactly (`settingSources: []`, `strictMcpConfig`, `Read(./**)`/`Edit(./**)`, sandbox `failIfUnavailable: true`, Bash network deny-all); pins are exact (SDK 0.3.251, CLI 2.1.251 via the SDK's linux-arm64 manifest — a justified mechanism change from a global install, not drift); the ENTRYPOINT matches the CI boot probe; the README runbook's connector/role names match `infra/terraform/microvm.tf` and the gateway path matches chat-api's `/v2/gateway/:conversationId`. No scope creep found (`MICROVM_IMAGE_NAME` override and the RDS CA bake serve this ticket's verification path). Out-of-scope tickets (#664 drain loop — merged in from main, #665, #668/#669, #670) treated as such.

Non-blocking spec findings:

1. **No in-flight latch on `/run`** (`src/index.ts`). `if (serving)` catches only a *completed* configure; a duplicate `/run` arriving mid-configure (platform retry-on-timeout) would start a second `configure()` — two relays/DB pools/drain loops. The at-most-once invariant survives (claiming is DB-gated), and the window is one DB round-trip wide, so this is robustness, not correctness. Relatedly, a failed `configure()` leaks the relay/DB pool it created before the sweep threw — bounded by the failed launch killing the VM. Storing the configure promise fixes both; suggest as a follow-up.
2. **CI trigger paths miss some context inputs**: `stage_microvm_image_context.sh` copies the root and every workspace member's `package.json` into the image context, but `microvm-image.yml` triggers only on `apps/in-vm-server/**`, `packages/{agent-db,live-text}/**`, `bun.lock`, and its own script paths. A root/other-member manifest change alters the context untested; usually co-travels with `bun.lock`, so mostly benign.

## Standards

No hard violations. Dockerfile bakes exactly the four documented non-secret env vars; secrets ride `runHookPayload` into the trusted process only; config docs updated in step with the code; the README's plain-`aws` snippets sit under its `AWS_PROFILE=mymemo` preamble; the drizzle/pglite invariant is unaffected. The resolver duplication from agentcore-runtime is documented-accepted (ADR-0034).

Judgement calls (non-blocking):

- **Tautological assertions** (CODING_STANDARDS.md) in `claude-code-executable.test.ts`: `expect(executable).not.toContain("musl")` can't fail once the preceding `toContain` passes, and that assertion mirrors the implementation's own template literal. The exec-verify half of the test is the real value; the string assertions are decoration.
- `/run` passes `await c.req.json()` unchecked into a handler typed `{ microvmId?; runHookPayload? }` (security.md's "request-body schemas are strict"). Mitigated: the caller is the JWE-authenticated platform and `envFromRunHookPayload` strictly validates the one load-bearing field, failing closed.
- `PORT` validation moved from tested `config.ts` into the untested entrypoint; its test was deleted rather than relocated.
- The RDS bundle digest is now pinned in both `scripts/deploy/stage_microvm_image_context.sh` and `apps/agentcore-runtime/Dockerfile` — two-file surgery on rotation; transient until AgentCore retires, but a one-line "deliberate duplicate" comment in the staging script would match how the resolver copy is annotated.
- `AppHandlers.smokeScriptPath` is config living in an interface named "handlers" — mild naming wrinkle.

## Live-verification gap (documented, not a blocker)

Four of the five #666 ACs (hand-launched VM reaches ready, nudge drains a Turn end-to-end, credential-free CLI env asserted in staging, bwrap engaging in-VM) still require operator credentials this session couldn't use. The offline substitutes (CI boot-contract check on the baked entrypoint, smoke.sh assertions, 80 unit tests) are real, and the README runbook they depend on now states the boot contract correctly. #666 should stay open past merge until the runbook is executed.

Summary: Spec — 2 non-blocking findings, worst is the missing `/run` in-flight latch. Standards — 5 judgement calls, worst is the tautological string assertions. Both prior blockers verified fixed.

APPROVED FOR MERGE

chengchao:
Rebased onto main (`56a400b`) as one squashed commit `dee50c0`, replacing the merge-commit history. Verified: `git diff 991085d dee50c0` is empty and the PR diff vs main is byte-identical to the reviewed state — the Ponytail NO FINDINGS and APPROVED FOR MERGE verdicts on `991085d` apply to this exact tree. CI re-runs on the new head.

Issue #654, opened 2026-08-31
Spec: /v2 chat on Lambda MicroVMs — Claude Agent SDK in-VM behind an egress gateway


chengchao:
Produced by [Wayfinder map: /v2 chat on Lambda MicroVMs](https://github.com/X-GPT/mymemo-agent/issues/644) — eight resolved decisions collapse into this spec. Each section links the resolution that owns its detail; this document is the buildable contract, the resolutions are the reasoning record. ADR-0034 (PR linked from the map) carries the trust-boundary stance; the same PR performs the `CONTEXT.md` glossary surgery.

**Amendments** (body already reflects them): [2026-08-31 — stock AI SDK v7 stream, one assistant UIMessage per Turn](https://github.com/X-GPT/mymemo-agent/issues/654#issuecomment-5482609743) · [2026-09-01 — no shell in the VM, Bash dropped](https://github.com/X-GPT/mymemo-agent/issues/654#issuecomment-5499996397).

## Goal

MyMemo chat runs on `/v2/*`: chat-api fronts **one persistent AWS Lambda MicroVM per Conversation** running the **Claude Agent SDK in-VM** (pinned `0.3.251`), with file tools confined to cwd, **no shell** (`Bash` denied — [ADR-0034 amendment](https://github.com/X-GPT/mymemo-agent/blob/main/docs/adr/0034-run-the-chat-loop-in-per-conversation-lambda-microvms.md#amendment-2026-09-01--no-shell-in-the-vm)), **all data-plane credentials confined to the trusted in-VM server process**, egress locked to a **credential-injecting gateway route in chat-api**, turns streamed as **UIMessages over the v1 Redis relay**, and Postgres as the work bus. v1 + AgentCore + E2B keep serving production until v2 cutover; their teardown is a named follow-on.

## The /v2 API surface

| Route | Contract |
| --- | --- |
| `GET /v2/conversations` / `POST /v2/conversations` / `PATCH\|DELETE /v2/conversations/:id` | carried over from v1 semantics (owner-scoped, exposure-gated creation, frozen Scope; deletion flow below) |
| `POST /v2/conversations/:id/messages` | submit a UIMessage; response is the standard **UIMessage SSE scoped to the submitted message's own turn** (silent keepalives while queued predecessors drain). No 409: a second POST is just the next queued row. Re-POST of the same client message id is a no-op (PK). |
[...]

Issue #610, opened 2026-08-27
Where does the framework-neutral tool catalog live and how does it adapt to both paths?


chengchao:
Part of #607

## Question

Where does the framework-neutral tool catalog live, and how does it adapt to both consumers?

`buildRunTools()` in `apps/agentcore-runtime/src/sdk/run-tools.ts` returns Claude Agent SDK `SdkMcpToolDefinition[]` whose handlers close over per-Run E2B clients, the frozen document scope, limits, and the audit binding. Stage 2 needs the same handlers as an AI SDK `ToolSet` for `new HarnessAgent({ tools })` in chat-api, without the two catalogs drifting (the feasibility research's `RunToolDefinition` split).

Decide:

- **Package boundary.** Where the neutral definitions and the bounded handlers (`bash-tool/`, `file-tools/`, `documents/`, E2B clients) live so chat-api can import them: a new workspace package, or exported from `apps/agentcore-runtime` (apps importing apps is not a pattern here). Respect the Drizzle dependency invariant in `docs/agents/development.md` and the module boundaries in `docs/agents/architecture.md`.
- **Neutral shape.** The `RunToolDefinition`-style type (name, description, JSON Schema or Zod input, `execute` returning the existing `{ content[], isError? }`) and the two thin adapters: → `SdkMcpToolDefinition` (Run path, unchanged behavior) and → AI SDK `tool()` (Harness path). What the AI SDK `execute` returns to the relay so Claude does not receive an MCP envelope inside an MCP result; how `isError` maps.
- **Binding.** The catalog's binding type once `runId` is optional/replaced on the Harness path (depends on the Workspace ticket's vocabulary).
- **Construction time.** `HarnessAgent` takes `tools` at construction and stage 1 constructs one agent at composition; per-Conversation closures need a per-turn agent (or a tool layer that resolves the Conversation from a per-call context). Pick one and state the cost.
- **Invariants to test.** The `experimental_sandbox` discard regression; a pin that the Harness `ToolSet` and `EXECUTOR_ALLOWED_TOOLS` are built from one list.

Output: package layout, the neutral type, adapter signatures, and the per-turn construction decision.

- **Output bounding (graduated from the map's fog after #608).** The bridge persists every host-tool output verbatim into the sandbox transcript and streams it as `tool-output-available`; the AI SDK adapter is therefore where output size is bounded and secrets from E2B responses are redacted. Decide the bound and where it lives (per-tool limits already exist for the Run path).

[...]

PR #688, opened 2026-09-01
feat(in-vm-server): doc tools as in-process MCP via shared @mymemo/document-tools (#665)

### LINKED ISSUE #665: v2: doc tools as in-process MCP
## Parent

## What to build

The three document tools — ListDocuments, SearchDocuments, LoadDocuments — served to the SDK loop as in-process MCP inside the trusted In-VM server, with Scope-enforced KB SQL and the docs cache materialized in the Workspace. Prefactor first: extract the portable tool and scoped-client logic out of chat-api into a workspace package consumed by both sides (chat-api's local path keeps working until teardown) — this also prevents a third copy of the frozen-Scope parsing (a deliberate duplicate already exists per #610). Load stays disk-only with a metadata-only result; re-Load refreshes the cached copy; per-call audit rows are written by the trusted process. The spawned CLI holds no KB credential — tools execute in the trusted process.

## Acceptance criteria

- [x] A local Turn can list, search, and Load real KB documents strictly within the Conversation's frozen Scope — `general`, `collection`, and `document` shapes each covered.
- [x] Load materializes content into the Workspace docs cache and returns metadata only; re-Load refreshes.
- [x] Out-of-scope access fails closed and never appears in results.
- [x] Audit rows written per call by the trusted process.
- [x] chat-api's existing consumers pass their tests against the extracted package — no logic fork.
- [x] The CLI env carries no KB credential (asserted).

## Blocked by

chengchao:
Closes #665. Parent: spec #654 (/v2 chat on Lambda MicroVMs).

## What this builds

**Prefactor — `packages/document-tools` (`@mymemo/document-tools`).** The portable pieces of chat-api's Harness document tools move to a shared workspace package, consumed by both chat-api's local chat path and the In-VM server, so no third copy of the frozen-Scope parsing exists (the Run path's `apps/agentcore-runtime/src/documents/` remains the one deliberate #610 duplicate):

- `client` — `parseFrozenScope` (structural row type, fail-closed on all three shapes), the scoped read-only KB client (`createScopedDocumentClient`: server-side scope narrowing, uniform `OUT_OF_SCOPE` for unknown *and* out-of-scope ids, bounded infra errors, audit-write failure fails the call closed), `createKbDb`.
- `tools` — the three handlers (`listDocuments`/`searchDocuments`/`loadDocuments`) with the code-constant caps (20 list / 8 search / 10 docs · 256 KiB · 1 MiB), the cursor codec, the `DocsCacheWriter` seam, plus newly shared `DOCUMENT_TOOL_NAMES` and `DOCUMENT_TOOL_DESCRIPTIONS` so the two consumers' catalogs cannot drift.
- `access-log` — `DocumentAccessEvent`/`DocumentAccessLog` and the Drizzle `PostgresDocumentAccessLog` writer.

Git records the moves as renames at 91–98% similarity — the extraction is mechanical, no logic fork. chat-api keeps only the Harness `ToolSet` adapter (`ai-chat/tools/harness-tools.ts`); its route, deps, and factory just re-point imports. The package mirrors agent-db's peer-dependency shape (`drizzle-orm` + `pg` peers, `@electric-sql/pglite` dev) so the single-Drizzle-instance invariant holds.

**The In-VM side — doc tools as in-process MCP (`apps/in-vm-server/src/doc-tools.ts`).** `createDocToolsServer` wraps the three tools in `createSdkMcpServer({ name: "mymemo-docs", alwaysLoad: true })`, wired into the confinement bundle: `mcpServers` carries exactly this server (`strictMcpConfig` already ignores everything else) and `allowedTools` gains the `mcp__mymemo-docs__*` names, pinned against the actually-built tool list so the two cannot drift.

- **Scope**: the Conversation's frozen Scope is read from the agent DB (`conversations` row) lazily on first tool call and cached only on success — Scope is immutable for the Conversation's lifetime, and a local server may start before its Conversation row exists. An unresolvable row fails the call closed with a fixed, model-safe message.
- **Turn attribution**: a shared `CurrentTurn` ref is set by `serveOneTurn` at claim and cleared in its `finally`; every `document_access_events` row is keyed by the in-flight Turn's user-message id (`run_id`), one row per list/search/per-document load, written by the trusted process through the shared Drizzle writer. A call with no Turn in flight fails closed rather than write an unattributable row. The ref survives #664's long-lived `query()` design (the MCP server is built once per VM).
- **Docs cache**: `LoadDocuments` materializes to `<WORKSPACE_DIR>/.mymemo/docs/<id>.md` through a plain local-FS writer (inside the VM the Workspace is the process's own disk — no sandbox RPC), returns metadata only (path, bytes, truncated), and a re-Load overwrites the cached copy.
- **Credentials**: new required `KB_DATABASE_URL` in `InVmConfig`, held by the trusted process only. The CLI env allowlist is untouched; the poisoned-env test now asserts `KB_DATABASE_URL` explicitly.

## Acceptance criteria

- **A local Turn lists, searches, and Loads real KB documents within the frozen Scope — all three shapes covered (verified live)**: compose Postgres (seeded KB) + Redis + OpenRouter, one server per Conversation, real `query()` turns:
  - *general*: model called `ListDocuments` (total 2) → `SearchDocuments` ("machine learning", 1 hit) → `LoadDocuments(doc-ml-intro)` → `Read` on the cached file → answered `TITLE=Intro to Machine Learning`; 3 audit rows keyed `msg-665-gen-1`.
  - *collection* (`demo-collection`): `LoadDocuments(["doc-mymemo-overview","doc-ml-intro"])` → answered `LOADED=doc-ml-intro FAILED=doc-mymemo-overview`; only the in-scope doc cached and audited.
  - *document* (summaryId 1001): `ListDocuments` → `DOCS=doc-ml-intro` only; `SearchDocuments("personal knowledge base")` — text present only in the out-of-scope doc — `HITS=0`.
- **Load materializes into the Workspace docs cache, metadata-only; re-Load refreshes**: live above, plus `doc-tools.test.ts` (file content asserted, result carries path/bytes and never the body, second Load overwrites).
- **Out-of-scope access fails closed and never appears in results**: live collection run above; unit tests pin the uniform message, no file write, no audit row, and (in the package tests) the unknown-id ≡ out-of-scope-id indistinguishability.
- **Audit rows written per call by the trusted process**: live rows shown above (`run_id` = Turn user-message id, `scope_type`/`scope_id` per shape); pglite tests insert through the real `PostgresDocumentAccessLog`.
- **chat-api's existing consumers pass their tests against the extracted package**: full chat-api suite green (174 tests) with the tools imported from `@mymemo/document-tools`; the moved tests run in the package (20).
- **The CLI env carries no KB credential (asserted)**: `query-options.test.ts` seeds `KB_DATABASE_URL` in the poisoned trusted env and asserts the exact allowlist-only key set plus an explicit `KB_DATABASE_URL === undefined`.

## Notes for review

- No system-prompt append here: the MCP tool descriptions are the model-facing contract, and the general-scope live turn shows stock prompting finds and uses the tools. Prompt shaping can ride a later ticket if wanted.
- The scoped client is constructed per call (closures only, no I/O) because the binding carries the per-Turn id while the MCP server is per-VM.
- `DOCUMENT_TOOL_DESCRIPTIONS` is newly shared between chat-api and the In-VM server — one implementation, one wording. The #610 stance (Run path deliberately separate) is unchanged and re-recorded in `docs/agents/chat-api.md`.
- `drizzle-orm` moves from dev to runtime deps in in-vm-server (the Scope read uses `eq`/`and`).
- Pre-existing, untouched: chat-api's `bunx tsc --noEmit` reports one `s3-artifact-download-signer.ts` error from a `@smithy/types` version fork in the lockfile — present on `main`, unrelated to this change.

## Verification

- `packages/document-tools`: 20 pass; `bunx tsc --noEmit` clean.
- `apps/in-vm-server`: 65 pass (new: doc-tools suite, allowedTools/mcpServers pins, Turn-ref tests); `bunx tsc --noEmit` clean.
- `apps/chat-api`: 174 pass.
- Full root suite (`bun run test`): all workspaces green; biome clean.
- Live end-to-end: the three scope-shape runs above (local compose + OpenRouter), seeded rows cleaned up afterwards.

https://claude.ai/code/session_01RUK9UyR1YzefzvptsFLXR3


chengchao:
## Review — #665 (doc tools as in-process MCP), diff `origin/main...0c106e6`

Two-axis review (Standards per AGENTS.md / docs/agents / CODING_STANDARDS.md; Spec per issue #665). Every finding below was verified against the code at 0c106e6.

**No documented-standard violations found.** Checked specifically: the Drizzle single-instance invariant (`drizzle-orm` promoted to `dependencies` in `apps/in-vm-server/package.json`, `@electric-sql/pglite` kept as a dev dep in both consumers; `packages/document-tools`' peer shape mirrors `agent-db`); the security boundary (`KB_DATABASE_URL` held by the trusted process only, pinned absent from the CLI env in `apps/in-vm-server/src/query-options.test.ts:129-135` against an env that does contain it — the allowlist test is not vacuous); `strictMcpConfig` retained with exactly the one in-process server; `alwaysLoad` confirmed real in SDK 0.3.251 types; docs (architecture/security/database/configuration/chat-api) updated coherently; `LoadDocuments` filenames guarded by `SAFE_DOCUMENT_ID` so no traversal out of the docs cache.

### Findings

1. **Minor (optional) — the tool input schemas are the one catalog still written twice.** The extraction shares `DOCUMENT_TOOL_NAMES` and `DOCUMENT_TOOL_DESCRIPTIONS` "so the catalogs cannot drift", but the zod input shapes are duplicated between `apps/in-vm-server/src/doc-tools.ts:177,184,193` and `apps/chat-api/src/features/ai-chat/tools/harness-tools.ts:71-93`. Exporting the raw shapes from `@mymemo/document-tools/tools` would serve both (SDK `tool()` takes the raw shape; the Harness path wraps it in `z.object(...)`). Not merge-blocking.

2. **Minor (optional) — `document` scope shape not driven through the new in-VM path.** `apps/in-vm-server/src/doc-tools.test.ts` exercises `buildDocTools`' scope resolution only under `general` and `collection`; the `document` shape is covered solely by the shared package tests (`packages/document-tools/src/client.test.ts:169`). AC1 names all three shapes; one added case would evidence it at the layer this PR introduces. (AC1's "real KB documents" half still needs the customary live-verification comment before ticking.)

3. **Note (no change requested) — a scope-refused Load leaves no audit row.** `fetch`'s scope guard in `packages/document-tools/src/client.ts` refuses before auditing, and `doc-tools.test.ts:341-362` pins zero audit rows for the out-of-scope case. This is pre-existing chat-api behavior carried over mechanically — the "no logic fork" AC mandates keeping it — but AC3+AC4 read together arguably want denied attempts in the ledger. Worth a deliberate call (follow-up ticket if desired), not a change in this PR.

4. **Note (merge prerequisite) — rebase onto #664 required.** The branch is based on 5e490d9 (pre-#664); main's drain loop conflicts in `apps/in-vm-server/src/index.ts` (a rebase is in progress). When resolving, preserve the `currentTurn` set-before-`query()` / clear-in-`finally` ordering from `turn-serving.ts` around #664's long-lived query so audit rows stay attributed to the right Turn across drained Turns — the shared-ref design in `doc-tools.ts` already anticipates this.

None of these requires a code change before merge: 1 and 2 are optional improvements, 3 is mandated by the no-fork AC, 4 is routine and already underway.

APPROVED FOR MERGE

chengchao:
Response to the review (all four points addressed; branch rebased onto main's #664 drain loop, head now `3a0fb80`):

1. **Input-schema sharing** — left as-is deliberately. #610 accepted schema duplication as the cheap half of the catalog (names + descriptions were the drift-prone parts and are now shared); two ponytail passes on this diff also weighed and skipped the shape export. Happy to fold it in if a third consumer ever appears.
2. **`document` scope at the in-VM layer** — adopted: `doc-tools.test.ts` now drives `buildDocTools` under the `document` shape too (summaryId → resolved KB id pinned in the search SQL and the audit row) in `3a0fb80`. The "real KB" half of AC1 was verified live before the PR opened — see **Verification** in the PR body: all three scope shapes ran end-to-end through real `query()` turns against the seeded compose KB (general: List/Search/Load + `Read` of the cached file; collection: in-scope doc loaded, out-of-scope doc refused; document: list pinned to the summary document, out-of-scope text unfindable).
3. **No audit row for a scope-refused Load** — confirmed as the deliberate carry-over the no-fork AC mandates (the Run path behaves the same way). Agreed it deserves a deliberate product call; proposed as a follow-up ticket ("audit denied document access attempts") rather than a divergence here.
4. **Rebase onto #664** — done. The conflict was confined to `index.ts` wiring; `currentTurn` keeps its set-after-claim / clear-in-`finally` ordering in `turn-serving.ts`, which #664's drain loop calls sequentially, and the long-lived Agent session reuses the same options bundle so the in-process MCP server persists across Turns as designed. The drain-loop tests now wire the (required) `currentTurn` dep; full suite green on the rebased head.

chengchao:
## Review — #665 (doc tools as in-process MCP), rebased head `3a0fb80`

Fresh two-axis review (Standards per AGENTS.md / docs/agents / CODING_STANDARDS.md; Spec per issue #665) of the rebased branch, base `origin/main` (56a400b). Every finding below was verified against the code at `3a0fb80`. All suites green locally (document-tools 20, in-vm-server 76, chat-api ai-chat 21), `tsc --noEmit` clean on `packages/document-tools` and `apps/in-vm-server`, Biome clean, chat-api carrying only the pre-existing `s3-artifact-download-signer` error already on main.

The prior round's four points are all resolved on this head: the `document`-scope handler case is present (`apps/in-vm-server/src/doc-tools.test.ts:192`), the audit-on-refused-Load carry-over and the #610 zod-shape duplication stand as deliberate, and the #664 rebase preserves the `currentTurn` set-after-claim / clear-in-`finally` ordering (`apps/in-vm-server/src/turn-serving.ts:85,165`).

### Finding 1 — Major (fix before merge): LoadDocuments materializes in the trusted process with no symlink guard

`apps/in-vm-server/src/doc-tools.ts:88-93` — the new `localDocsCacheWriter`:

```ts
async writeTextFile({ path: filePath, content }) {
	await mkdir(path.dirname(filePath), { recursive: true });
	await writeFile(filePath, content, "utf8");
},
```

runs in the **trusted** In-VM process (unsandboxed, holding `KB_DATABASE_URL` / `AGENT_DATABASE_URL`) and writes into `<workspaceDir>/.mymemo/docs/<id>.md` — a tree the **untrusted** CLI freely populates (`Edit(./**)` plus sandboxed Bash with cwd write, `ln -s` included; `apps/in-vm-server/src/query-options.ts:86-93`). `fs.writeFile` (flag `w`) follows a symlink in the final component, and `mkdir(..., { recursive: true })` silently accepts a symlinked parent. So the model can pre-plant `.mymemo/docs/<id>.md` — or `.mymemo/docs` itself — as a symlink to any path, and the next LoadDocuments makes the trusted process follow it and write largely model-influenceable KB content (with `O_TRUNC`) outside the Workspace. That is a confused-deputy filesystem write across the exact process boundary this feature exists to defend (ADR-0034; docs/agents/security.md "a process boundary is the trust boundary").

This is new to #665's wiring, not a suppressed #610/no-fork carry-over: both sibling paths route this write **through the untrusted sandbox** — Harness via `experimental_sandbox` (`apps/chat-api/src/features/ai-chat/tools/harness-tools.ts:96-107`), the Run path via `context.sandbox.writeFile` into E2B (`apps/agentcore-runtime/src/documents/load-documents-tool.ts:159`) — so neither needed a guard; the write never carried trusted authority before. `SAFE_DOCUMENT_ID` (`packages/document-tools/src/tools.ts:165`) blocks traversal in the filename but not a pre-planted symlink at the target, so it does not cover this.

The fix is small and belongs in `localDocsCacheWriter` (the shared `loadDocuments` needs no change): unlink the leaf then open with `wx`/`O_NOFOLLOW` (defeats the leaf-symlink case), plus a realpath-containment check that the resolved parent stays under the Workspace (defeats the parent-symlink case) — with one test that plants a symlink and asserts the write is refused. Blast radius is one ephemeral single-tenant VM, but the reachable targets — the trusted server's own files and the image's managed-policy tier that `settingSources: []` still reads — are precisely what the boundary protects, so this is worth closing here rather than deferring.

### Standards axis — no documented-standard violations

Checked and satisfied: the Drizzle single-instance invariant (`drizzle-orm` promoted to dependencies in `apps/in-vm-server/package.json` with `@electric-sql/pglite` kept as a dev dep; `packages/document-tools` lists both peer+dev), the security boundary pins (`query-options.test.ts:129-134` asserts `KB_DATABASE_URL` absent from the CLI env against a poisoned process env, so the test is not vacuous), domain terminology, and doc coherence (security/database/configuration/architecture/chat-api all updated in-diff). Two verified minor, non-blocking notes:

1. **Stale module comment** — `packages/document-tools/src/client.ts:7-8` still says the client is "bound once per turn instead of per call", but the In-VM consumer (`doc-tools.ts` `withClient`) now constructs a scoped client per tool call. One-line comment fix.
2. **Test-helper duplication** — the `fakeKb` recorder exists verbatim twice (`packages/document-tools/src/client.test.ts:18`, `apps/in-vm-server/src/doc-tools.test.ts:62`). Test-only; fold into `@mymemo/test-support` whenever either file is next touched.

### Spec axis — all six ACs otherwise trace end to end

KB credential confined to the trusted process (`config.ts:71`, `index.ts:31`), the in-process server injected under `strictMcpConfig` with the allowlist pinned to the built tools (`doc-tools.test.ts` "the allowlist pin"), metadata-only Load + re-Load refresh and out-of-scope fail-closed covered at the handler level, per-call audit rows attributed via the in-flight-Turn ref, and AC6 asserted. AC1's "real KB documents" half was verified live per the PR body. The zod input shapes in `doc-tools.ts:177,184,193` are a second hand-copy of the Harness shapes — same drift class as the shared catalog, but schema duplication was accepted on #610 and reaffirmed last round; not re-litigated.

Finding 1 is the only change requested.

CHANGES REQUIRED

chengchao:
All three findings from the CHANGES REQUIRED review are addressed in `99763ae`:

1. **Major — confused-deputy write through a planted symlink**: fixed. `createLocalDocsCacheWriter` now (a) realpath-pins the resolved cache directory under the realpath of the Workspace root after `mkdir` — a planted directory symlink fails the call closed with a fixed model-safe message; (b) replaces the leaf (`rm force`, which unlinks a symlink rather than following it) and creates it with `O_EXCL` (`wx`), which never follows a symlink, so a re-planted leaf fails closed instead of writing elsewhere. Both escape shapes are pinned by new symlink-planting tests ("refuses to write through a planted directory symlink out of the Workspace" — the outside directory stays empty; "replaces a planted leaf symlink instead of writing through it" — the outside target stays untouched and the cache entry becomes a regular file).
2. **Stale comment** in `packages/document-tools/src/client.ts` — reworded to "bound to one turn's binding at construction".
3. **`fakeKb` duplication** — left as-is: test-local fakes are the repo's convention, and a prior ponytail pass judged the ~15-line duplication cheaper than a cross-package testing export; will fold into `@mymemo/test-support` if a third copy ever appears.

Full in-vm-server suite 78 pass, tsc and Biome clean.

chengchao:
## Review — #665 (doc tools as in-process MCP), head `99763ae`

Fresh two-axis review (Standards per AGENTS.md / docs/agents / CODING_STANDARDS.md; Spec per issue #665) at head `99763ae64cb1d12c68e954c9e6ef903e470db8f9`, base `origin/main` (56a400b). Every finding verified against the code. Suites green (document-tools 20, in-vm-server 78, chat-api ai-chat 21), `tsc --noEmit` clean on document-tools and in-vm-server (chat-api carries only the pre-existing `s3-artifact-download-signer` error already on main), Biome clean, Drizzle invariant satisfied.

### The prior round's Major (symlink-planted docs-cache write) — fixed, verified correct and sufficient

`apps/in-vm-server/src/doc-tools.ts:95-109` (`createLocalDocsCacheWriter`):

- **Parent shape closed**: `realpath` after `mkdir` collapses any depth of planted directory symlink; containment compares realpaths on both sides (handles macOS `/var` → `/private/var`) and `realDir !== realRoot && !realDir.startsWith(realRoot + path.sep)` avoids the `/workspace` vs `/workspace-evil` prefix pitfall. The write then targets `path.join(realDir, basename)`, not the caller's logical path.
- **Leaf shape closed**: `rm(force)` unlinks a leaf symlink without following it; `writeFile(..., { flag: "wx" })` is `O_CREAT|O_EXCL`, which never follows a symlink and fails `EEXIST` on any re-planted leaf (dangling included). A directory planted at the leaf makes `rm` throw — also fail-closed. `SAFE_DOCUMENT_ID` keeps `basename` traversal-free.
- **Tests pin both shapes** (`apps/in-vm-server/src/doc-tools.test.ts:270-307`): each fails against the pre-fix flag-`w` writer (the directory test's `readdir(outside)` would be non-empty; the leaf test's target bytes would change). Ran the suite at head: pass.

Two named residuals, **neither blocking**: (a) TOCTOU — the check→open window rebuilds paths from strings, so a background race that swaps a validated parent for a symlink after `realpath` could still redirect the create; unclosable in Node (no `openat`), and `O_EXCL` caps a race win at creating a new file, never truncating an existing one — worth a one-line comment naming the ceiling, no code change requested. (b) `mkdir` at `doc-tools.ts:98` runs before the containment check, so a planted *dangling* symlink can make the trusted process create an empty directory outside the Workspace before failing closed — no content escapes; fine to leave.

### Finding 1 — Major (fix before merge): chat-api's Docker image no longer boots — `packages/document-tools` is never copied in

`apps/chat-api/Dockerfile:34-37` (prerelease), `:43-46` (local), `:54-65` (release) copy only `packages/agent-db` and `packages/live-text` sources. This PR makes chat-api import the new workspace package at boot — `apps/chat-api/src/deps.ts:2-5` (`@mymemo/document-tools/access-log`), reached unconditionally from `src/index.ts` via `createDeps` — and the package's exports map to `./src/*.ts` (`packages/document-tools/package.json:5-9`), resolved through the workspace symlink in the copied `node_modules`. In the image that symlink dangles: both the `release` image and compose's chat-api service (`target: local`, `compose.yaml`) fail module resolution at startup. That breaks the spec's own guardrail — "chat-api's local path keeps working until teardown" — and the Dockerfile's documented contract (lines 60-64). Fix is three lines: `COPY packages/document-tools/ packages/document-tools/` in `prerelease`, mirrored as `COPY --from=prerelease ...` in `local` and `release`.

(Checked the other consumer: the MicroVM image (`infra/microvm-image/Dockerfile`) is still the #661 placeholder and does not yet bundle in-vm-server, so chat-api's Dockerfile is the only instance.)

### Everything else — clean

- Standards: no documented-standard violations. Drizzle invariant holds (`drizzle-orm` in in-vm-server `dependencies`, pglite stays dev; document-tools pairs peer+dev). `KB_DATABASE_URL` confined to the trusted process and pinned absent from the CLI env against a poisoned process env (`query-options.test.ts`). Docs updated coherently. The in-vm-server's direct `.from(conversations)` read matches the existing app-side store pattern (e.g. `apps/chat-api/src/features/run-store/run-store.ts:86`), and the `sandbox: DocsCacheWriter` field name is verbatim from main's rename source — both fine under the mechanical extraction.
- Spec: all six ACs trace — scope shapes covered at the handler level (`doc-tools.test.ts:189-213`) with the full matrix in the package tests, metadata-only Load + re-Load refresh (`:232-266`), out-of-scope fail-closed with no file and no audit row (`:309-330`), per-call audit rows attributed via `currentTurn` set-after-claim / clear-in-`finally` (`turn-serving.ts:85,165`), no logic fork (the package diff is rename + identifier churn only), CLI env asserted credential-free. No scope creep. Round 2's minor notes (stale client.ts comment reworded; fakeKb duplication declined per convention) are settled.

Finding 1 is the only change requested — it's outside the diff's files but caused by it, and it takes the deployment artifact down at boot.

CHANGES REQUIRED

chengchao:
The round-2 findings are addressed in `2a1623b`:

1. **Major — chat-api image missing `packages/document-tools`**: fixed. The package is now copied in all three stages (`prerelease`, `local`, `release`), mirroring agent-db, with a comment on the release copy explaining why it must ship. Verified beyond the build: `docker build --target release` then `docker run` boots past module resolution and fails on the expected `AGENT_DATABASE_URL is required` config assertion — the workspace symlink resolves.
2. **Non-blocking residuals on the docs-cache writer**: adopted as a ceiling note — the writer's doc comment now names the unclosable-in-Node parent-swap TOCTOU (O_EXCL caps a race win at creating a new file, never truncation/overwrite) and the mkdir-before-check empty-directory side effect of a dangling planted symlink.

in-vm-server suite green (78), Biome clean.

chengchao:
## Review — #665 (doc tools as in-process MCP), head `2a1623b`

Fresh two-axis review (Standards per AGENTS.md / docs/agents / CODING_STANDARDS.md; Spec per issue #665) at head `2a1623b601defe8eee08fdcb48a2cf6f03b7b455`, base `origin/main` (56a400b). Every finding below was verified against the code; findings that did not survive verification are listed at the end rather than reported.

Suites green at this head: document-tools 20, in-vm-server 78, chat-api ai-chat 21. `tsc --noEmit` clean on `packages/document-tools` and `apps/in-vm-server`; chat-api reports only the pre-existing `s3-artifact-download-signer.ts(45,18)` `@smithy/types` fork error already on main. Biome clean across all 88 touched files.

Round 2's two blockers are fixed and re-verified here: `createLocalDocsCacheWriter` (`apps/in-vm-server/src/doc-tools.ts:95-115`) realpath-pins the parent and creates the leaf with `wx`, with both escape shapes pinned by symlink-planting tests (`doc-tools.test.ts:270`, `:291`); the chat-api Dockerfile copies `packages/document-tools` in all three stages (`apps/chat-api/Dockerfile:36`, `:46`, `:66`).

### Finding 1 — Minor (docs, non-blocking): `security.md`'s KB-locality sentence is stale after the extraction

`docs/agents/security.md:26` still reads:

> …because chat-api executes the document tools itself; **only `apps/chat-api/src/features/ai-chat/tools/` touches the KB**, and chat-api holds no E2B credential.

After this PR the KB pool and every scoped SQL statement live in `packages/document-tools/src/client.ts`; `ai-chat/tools/` keeps only the Harness `ToolSet` adapter, and chat-api's sole KB call site is now `apps/chat-api/src/features/ai-chat/harness-chat-agent.ts:82` (`createKbDb`) importing from `@mymemo/document-tools/client`. The diff corrected the sibling statements — `docs/agents/chat-api.md:65-69`, `docs/agents/architecture.md`, and `security.md:30` itself — but not this one, and `security.md` is the trust-boundary reference other work reads first. The substantive claim (production `ApiConfig` holds no KB credential; only the local `HarnessConfig` does) is unchanged and still correct — only the "which directory touches the KB" pointer is wrong. One-sentence fix: point it at `@mymemo/document-tools` with `ai-chat/tools/` as the adapter.

### Two optional nits (no change requested)

1. `apps/in-vm-server/src/doc-tools.ts:80-84` — `toCallToolResult(result: object | ToolFailure)` collapses to `object`, which is what forces the `as ToolFailure` cast one line down. Typing the parameter as the handlers' actual return union would drop the cast. Cosmetic.
2. `apps/in-vm-server/src/doc-tools.ts:216-228` — `createDocToolsServer` (including `alwaysLoad: true`) has no automated coverage; `query-options.test.ts:81` injects `{} as McpSdkServerConfigWithInstance`, so no test constructs the real SDK MCP server. It is an 8-line wrapper whose naming contract is pinned by the allowlist test and whose wiring was exercised by the live runs in the PR body — a test here would mostly assert the SDK's own behavior. Fine as is.

### Checked and dropped — recorded so they are not re-raised

- **`Read(./**)` vs. the dot-prefixed `.mymemo/docs` cache.** Worth asking, since a `**` that skips dotfiles would hand the model paths it cannot open. It does not hold: the CLI bundle (`claude-agent-sdk-darwin-x64@0.3.251`) sets `dot: true` on every one of its five glob-option sites and contains no `dot:false`/`dot:!1` anywhere, and the PR body's general-scope live run narrates a successful `Read` of the cached file. Not a defect.
- **`document`-shape Load/List coverage.** `packages/document-tools/src/client.test.ts:169-194` already drives the document scope through resolve → search → list → `fetch("kb-9")` admit → out-of-scope `fetch` refusal; the handler-level case is `doc-tools.test.ts:200`.
- **No in-vm-server compose service / no automated real-KB test.** `in-vm-server` has never had a compose service (0 occurrences on `origin/main`), so this PR regresses nothing; AC1's "real KB documents" half is the live verification in the PR body, and the automated tests use a fake KB by design.
- **`sandbox: DocsCacheWriter` field name; `frozenScope()`'s direct `.from(conversations)` select.** Both cleared last round — the field name is verbatim from the rename source under the no-fork AC, and the direct select matches the app-side store pattern (`apps/chat-api/src/features/conversation-store/postgres-conversation-store.ts:51`, `run-store.ts:86`).
- **The `mcp__<server>__<tool>` allowlist assertions.** Partly derived on both sides, but the `mcpServers` key is the same `DOC_TOOLS_SERVER_NAME` constant, so the prefix is correct-by-construction, and `query-options.test.ts:105-107` pins the three tool names literally. Not tautological in the sense CODING_STANDARDS.md forbids.

### Both axes otherwise clean

**Standards** — no documented-standard violations beyond Finding 1. Drizzle single-instance invariant satisfied (`apps/in-vm-server/package.json` promotes `drizzle-orm` to `dependencies` and keeps `@electric-sql/pglite` as a dev dep; `packages/document-tools/package.json` carries both), with no casts around dual-instance errors. New tests are load-bearing: `turn-serving.test.ts:448-477` fails if `currentTurn` is not set or not cleared, `config.test.ts:40` fails if the assert is dropped, `query-options.test.ts:112` fails on an extra or missing MCP server, and `doc-tools.test.ts` drives real handlers against PGlite.

**Spec** — all six ACs trace. Scope-enforced KB SQL and the docs cache in the trusted process; `KB_DATABASE_URL` confined to it and pinned absent from the CLI env against a poisoned process env (`query-options.test.ts:129-135`); metadata-only Load with a refreshing re-Load; out-of-scope fail-closed with no file and no audit row; per-call audit rows attributed via the set-after-claim / clear-in-`finally` Turn ref (`turn-serving.ts:85`, `:165`); the extraction is rename + identifier churn only. No scope creep — the "no Turn is being served" guard and `alwaysLoad: true` each trace to a spec requirement.

Nothing here requires a code change before merge; Finding 1 is a one-sentence doc correction that can ride this PR or a follow-up.

APPROVED FOR MERGE

chengchao:
## Review — #665 (doc tools as in-process MCP), head `d50b763`

Fresh two-axis review (Standards per AGENTS.md / docs/agents / CODING_STANDARDS.md; Spec per issue #665) at head `d50b76311deb26d4329d7fb3c1e102b2f45edae0`, base `origin/main` (56a400b). Every candidate finding was verified against the code; what did not survive verification is recorded below rather than reported.

**Delta from the round-3 approval**: `git diff 2a1623b..HEAD` is exactly the one-sentence `docs/agents/security.md:26` correction that round requested — and the new sentence's claim checks out: chat-api's sole KB construction site is `createKbDb(config.KB_DATABASE_URL)` at `apps/chat-api/src/features/ai-chat/harness-chat-agent.ts:82`, with `ai-chat/tools/` holding only the Harness adapter.

**Re-verified at this head**: document-tools 20 pass, in-vm-server 78 pass, chat-api ai-chat 21 pass; `tsc --noEmit` clean on `packages/document-tools` and `apps/in-vm-server` (chat-api carries only the pre-existing `s3-artifact-download-signer` `@smithy/types` error already on main); Biome clean across the 88 touched files; Drizzle single-instance invariant satisfied (`drizzle-orm` in in-vm-server `dependencies`, `@electric-sql/pglite` kept as a dev dep in both consumers). The round-2/3 fixes still hold: the symlink-hardened `createLocalDocsCacheWriter` with both escape shapes pinned (`apps/in-vm-server/src/doc-tools.test.ts:270`, `:291`), and `packages/document-tools` copied in all three chat-api Dockerfile stages (`apps/chat-api/Dockerfile:36`, `:46`, `:66`).

### Verified findings requiring a code change

None.

### One verified nit (optional, no change requested)

`apps/in-vm-server/src/turn-serving.ts:85` — `deps.currentTurn.turnId = claimed.messageId` is assigned before the `try` whose `finally` (`:165`) clears it, and `deps.relay.openPublisher(:88)` sits between; `openPublisher` has two contracted synchronous throws (`relay_closed`, invalid turn id), and the drain loop catches a `serveOneTurn` throw and continues (`drain-loop.ts:92-100`), so such a throw would leave the ref stale. Traced the consequence and it is not reachable as misattribution: audit rows are written only by tool handlers, handlers run only inside a `query()` iteration, and every `query()` invocation (`:115`) is strictly preceded by a fresh assignment at `:85` in the same call — the stale window contains no `query()` and therefore no tool calls, and both throw triggers are practically unreachable (the relay closes only at shutdown; the turn id is a DB-sourced UUID). Moving the assignment just inside the `try` would be a one-line defense-in-depth hardening that preserves the reviewed set-before-`query()` ordering; fine to leave as is or fold into any later touch of this file.

### Checked and dropped

- The `mcpServers` identity assertion in `query-options.test.ts:112` as a tautological-test breach — the "exactly one key" half carries the drift value, and the stub-server arrangement was adjudicated (the allowlist drift pin lives in `doc-tools.test.ts:132`).
- The hardcoded `tool("ListDocuments", ...)` literals vs `DOCUMENT_TOOL_NAMES` in `doc-tools.ts:196-215` — deliberate, and exactly what "the allowlist pin" test exists to hold together.
- The handler-level scope matrix being one-op-per-shape — the full op-by-shape matrix lives in `packages/document-tools/src/client.test.ts`, settled in round 1.
- `frozenScope()`'s cache-forever-after-first-success — correct because a Conversation's Scope is immutable for its lifetime, and only a successful read is cached (missing-row-then-recover is tested at `doc-tools.test.ts:173`).
- All previously adjudicated items (the mechanical extraction, the #610 duplicate incl. zod shapes, fakeKb duplication, peerDependencies shape, no-audit-row-on-refused-Load, the named TOCTOU ceilings, the `object | ToolFailure` cast, Read(./**) dotfiles) — none re-raised, none regressed.

### Both axes clean

**Standards** — no documented-standard violations. Security boundary intact: `KB_DATABASE_URL` confined to the trusted process (`config.ts:71`, `index.ts:32`) and pinned absent from the CLI env against a poisoned process env (`query-options.test.ts:129-135`); `strictMcpConfig` with exactly the one in-process server; docs (security/configuration/database/architecture/chat-api) all coherent with the code at this head.

**Spec** — all six ACs trace end to end: in-process MCP under `strictMcpConfig` with the drift-pinned `mcp__mymemo-docs__*` allowlist; all three scope shapes covered at the handler level with the full matrix in the package tests (AC1's "real KB" half verified live per the PR body); metadata-only Load with refreshing re-Load; out-of-scope fail-closed with no file and no audit row; per-call audit rows attributed via the set-after-claim / clear-in-`finally` Turn ref (clears on failure too, tested); no logic fork; CLI env asserted credential-free. No scope creep.

APPROVED FOR MERGE

Issue #692, opened 2026-09-01
Sandbox-mode Bash is unusable in the MicroVM: bwrap cannot mount /proc


chengchao:
Found live while running #666's verification runbook (spec #654, ADR-0034).

## What happens

A real Turn served from inside a real MicroVM completes, but **every Bash tool call fails**:

```
bwrap: Can't mount proc on /newroot/proc: Operation not permitted
```

Three calls, three identical failures (durable history on the verification Conversation). The model correctly diagnosed it as sandbox infrastructure, not a permission denial, and gave up.

## Why the smoke check said PASS

`infra/microvm-image/smoke.sh` ran only `bwrap --unshare-all --ro-bind / / --dev /dev true` — namespace creation. Claude Code's sandbox goes further. Per [`@anthropic-ai/sandbox-runtime`](https://github.com/anthropic-experimental/sandbox-runtime) (the engine behind sandbox-mode Bash), Linux sandboxing is two-stage:

1. an **outer bwrap layer** — filesystem/network/PID namespaces, bind mounts;
2. an **inner apply-seccomp layer** — nested user+PID+mount namespaces that **remounts `/proc`** to apply Unix-socket blocking via seccomp BPF.

Step 2 is what fails. Mounting procfs inside a user namespace is refused unless the host's `/proc` is "fully visible" — the masked-`/proc` rule that also blocks unprivileged Docker containers. The Lambda MicroVM runtime evidently presents a masked `/proc` to the application.

**The #646 probe carried the same gap** (`RESULT bwrap PASS` came from the namespace-only check), which is how "bubblewrap works at DEFAULT capabilities" entered the spec and ADR-0034 as settled fact. It is not settled: namespaces work, the sandbox does not.

PR #691 fixes the check (adds `bwrap --proc /proc` as `bwrap-proc`) so this can never pass silently again.

## Security posture: closed, not open

No command ran unsandboxed. `failIfUnavailable: true` + `allowUnsandboxedCommands: false` held — bwrap was present so the sandbox was deemed "available", the sandbox then failed at setup, and the tool call errored. The confinement invariant is intact; the Bash **capability** is unusable.

## Candidate remedies (untested)

[...]

Issue #651, opened 2026-08-30
What implements the egress gateway, and how are upstream credentials and spend governed?


chengchao:
Part of #644.

## Question

What implements the egress gateway, and how are upstream credentials, attribution, and spend governed?

The settled invariant: the VM's security groups allow outbound to exactly {RDS, Redis, gateway}; the gateway is the single internet door, forwarding allowlisted upstreams (OpenRouter first) with per-upstream credential injection. The VM already targets it via `OPENROUTER_BASE_URL` — config, not code. Open:

- **Implementation**: a dedicated minimal service (ECS) vs. a route inside chat-api vs. an off-the-shelf LLM proxy — judged on streaming passthrough, blast radius, and ops weight.
- **Key custody**: OpenRouter key from Secrets Manager to the gateway only; never in VM env, image, or checkpoint.
- **How the VM authenticates to the gateway**: is SG-source restriction alone enough (only VMs can reach it), or per-VM/per-Conversation credentials for attribution?
- **Attribution and spend**: per-Conversation usage tagging (the future hook for the token-charging map [#545](https://github.com/X-GPT/mymemo-agent/issues/545)), OpenRouter spend caps, and what the gateway logs.
- **Streaming passthrough**: SSE from OpenRouter through the gateway without buffering.

chengchao:
## Resolution — the gateway is a route in chat-api; egress full-routing verified live

### What implements the gateway

**A streaming passthrough route inside chat-api** — not a new service. chat-api is already in the VPC, always-on, trusted, and holds `chat_api_secrets`; the VM reaches it through the egress connector → chat-api's SG, never the public internet; coupling is a non-cost (if chat-api is down, chat is down). Streaming is a plain duplex `fetch`. Rejected: a dedicated ECS proxy (a deployable + pipeline + alarms for ~50 lines — the anti-pattern this map keeps refusing) and an off-the-shelf LLM proxy (a whole product for one upstream). **Recorded consequence**: production chat-api gains the OpenRouter secret — a deliberate revision to the v1 rule that kept provider creds out of it (that rule served the Run path's shape; "single internet door" is chat-api's trust class now).

### How the VM authenticates to the gateway

[...]

Issue #646, opened 2026-08-30
Probe a live MicroVM: SDK loop, sandbox-mode Bash, egress lockdown, suspend and rehydrate


chengchao:
Part of #644.

## Question

Stand up one probe MicroVM and run the feasibility checklist — does the design's load-bearing physics hold?

The [research brief](https://github.com/X-GPT/mymemo-agent/blob/research/aws-lambda-microvms/docs/research/aws-lambda-microvms-2026.md) established what's documented; this ticket measures what isn't:

1. The **Claude Agent SDK loop runs in-VM** (AWS samples only prove the CLI).
2. **Unprivileged user namespaces / bubblewrap** work in the guest kernel → sandbox-mode Bash actually sandboxes (the single most load-bearing unknown).
3. **Egress lockdown**: VPC egress connector into private subnets with no NAT — internet-bound traffic actually dies; RDS + ElastiCache reachable through the connector; their DNS names resolve (UDP-block quirk).
4. The settings bundle from the SDK-confinement ticket holds live: cwd-escape reads/writes denied, config-path writes denied, transcript intact.
5. **Suspend/resume** preserves the SDK process and transcript; measure resume latency on our real image.
6. The **S3 checkpoint/rehydrate cycle** (per the AWS Claude Code sample) round-trips `~/.claude` + workspace.
7. **SSE streams** through the authenticated per-VM endpoint to an external caller holding a JWE token.
8. Whether **suspended VMs count against the regional memory quota** (observe or ask support).

Asset: probe scripts + measured results on a research branch, linked from the resolution.

chengchao:
## Probe harness ready — live run blocked on tooling

The invariant probe kit is committed on `research/microvm-probe` (`docs/research/microvm-probe/`): a confinement image (SDK 0.3.251 + the root-owned managed-settings policy tier from `research/agent-sdk-confinement`), in-VM measurements (`probe.sh`), a lifecycle-hook + SSE server (`probe-server.mjs`), and a cheapest-first run order with the exact `lambda-microvms` / `lambda-core` command shapes ([README](https://github.com/X-GPT/mymemo-agent/blob/research/microvm-probe/docs/research/microvm-probe/README.md)).

**Two blockers before a live run — both need the human:**
1. **Tooling gap.** The local AWS CLI (2.34.23) ships no `lambda-microvms` service and AWS doesn't document a minimum version. Needs a CLI upgrade, an SDK driver (API version 2025-09-09), or a preview plugin.
[...]

PR #693, opened 2026-09-01
feat(in-vm-server): no shell in the VM — deny Bash and remove the sandbox machinery

chengchao:
Sandbox-mode Bash cannot start in the MicroVM, so the shell is denied outright and everything that existed to serve it is removed. This is the settled answer, not a holding pattern.

**Merge order: land #691 with this PR or after it — never before.** #691 removes the image side: bubblewrap and socat, the bwrap smoke check, and the policy tier's own `sandbox` block. Taking #691 first would strip bubblewrap while `main` still allows Bash under `failIfUnavailable: true`, leaving the tool demanding a sandbox that cannot be built. In the other direction the gap is harmless: until #691 lands, the image merely configures a sandbox for a shell that can no longer be invoked. The ADR amendment records this.

## The finding

During #666's live verification on a real Lambda MicroVM, every Bash tool call failed:

```
bwrap: Can't mount proc on /newroot/proc: Operation not permitted
```

Claude Code sandboxes Linux commands through `@anthropic-ai/sandbox-runtime`, which builds a nested layer that remounts `/proc`. Mounting procfs inside a user namespace is refused whenever the host's `/proc` is not "fully visible" — the masked-`/proc` rule that also blocks unprivileged containers — and the MicroVM runtime presents exactly that. The failure is closed, not open: nothing ran unsandboxed, the Bash tool was simply unusable.

## Why deny rather than weaken

Unsandboxed Bash inherits the VM's network, which hands the untrusted CLI:

1. **IMDS**, and with it the execution role's `conversations/*` S3 scope — the residual ADR-0034 accepted *because the in-VM process boundary contained it*. Without the sandbox, one Conversation's agent could read, overwrite, or delete every other Conversation's checkpoint.
2. **Unbounded model spend** on the real OpenRouter key: the gateway token is in the CLI env by design, and rate limiting on that path is deferred.
3. **An exfiltration channel** via the VPC resolver's public DNS and via the gateway itself.

None of those depend on unknowns. `enableWeakerNestedSandbox` (sandbox-runtime's Docker-compatible mode) was rejected on the same grounds — its own documentation says it considerably weakens isolation.

A shell was never a MyMemo requirement; it arrived as a Claude Code default and the spec then built confinement machinery around it. The agent keeps cwd-scoped `Read`/`Edit`/`Grep`/`Glob` and the in-process document tools (#665), which is what the product needs. #692 is closed as a documented limitation rather than a re-enable ticket.

## What changes

- `Bash`, `BashOutput`, `KillShell` move into `disallowedTools` — an explicit, testable denial rather than mere omission from the allowlist. What actually closes them, and every other tool the SDK ships, is `permissionMode: "dontAsk"` against the exact allowlist: anything unnamed is denied. The deny entries are defence in depth.
- The `sandbox` settings block and its dedicated test are **removed**: with no tool dispatching commands they guarded a code path that cannot execute. Keeping them as a guard against a careless future re-enable was rejected as speculative — the ADR now states that restoring a shell means restoring the bundle and remaking the security case.
- **ADR-0034 is amended.** Its decision body and its "measured facts this decision rests on" both rested on the false claim; the amendment records the mechanism and the rejected alternatives, and the measured fact is marked corrected in place rather than deleted, per the repo's ADR practice.
- `docs/agents/security.md` and the `CONTEXT.md` glossary follow, and stale "sandboxed Bash" references are swept from code comments, `.env.example`, and `docs/agents/configuration.md`.

## Verification

93 tests pass, `bunx tsc --noEmit` clean, biome clean, CI green. Eight ponytail review rounds (seven with findings, all applied) plus a code review.

https://claude.ai/code/session_01FBtkhfv1sRxGbXF9G8xz37


chengchao:
## Independent review — `feat/drop-bash-tool` @ `63c760c` vs `origin/main` (`5561dcc`)

Two-axis review (Standards / Spec). Verified locally: `bun test apps/in-vm-server` → **93 pass, 0 fail**; `tsc --noEmit -p apps/in-vm-server` → clean; `biome check` on all five changed TS files → clean.

The core decision is right and the code half is correct. Three things need fixing before this merges — all of them documentation-of-record, which is precisely what this PR exists to get right.

---

### The good part first

- The deny is **correctly layered**, not just cosmetic. `permissionMode: "dontAsk"` already makes an unlisted tool a terminal denial, and deny-beats-allow means the root-owned managed tier cannot re-open it. `Bash` / `BashOutput` / `KillShell` is the complete shell set — no other command-spawning built-in survives, `Task` / `SlashCommand` / `NotebookEdit` / `Write` are all absent from `allowedTools`, `settingSources: []` closes the hook route, and `strictMcpConfig` + one in-process server closes MCP.
- **No functional hole from dropping the shell.** ADR-0030 sent filename discovery to `Bash` with `rg --files`, but `/v2` keeps `Glob` in `allowedTools`, so discovery still works. (ADR-0030 is scoped to the AgentCore/E2B Run path that ADR-0034 already supersedes — no conflict, just worth knowing.)
- **No scope creep.** Every edit traces to the decision. The `sandbox` block and its test were genuinely dead once nothing dispatches commands, and the code-comment/doc sweep (`app.ts`, `config.ts`, `doc-tools.ts`, `.env.example`, `configuration.md`, `security.md`, `CONTEXT.md`) is complete on this PR's side.
- The rejection of unsandboxed Bash is argued from the right place: the `conversations/*` checkpoint scope is a residual ADR-0034 accepted *because* the process boundary contained it, so removing the containment reopens it. That reasoning belongs in the ADR and it landed there.

---

## Blocking

### 1. The PR title and body no longer describe this branch

This repo squash-merges, so the body becomes the permanent commit message. Right now it contradicts the code it would ship:

- **Title**: "deny the Bash tool **until the sandbox works**" — but `1c3b671` deliberately reframes this as settled, and the ADR now says restoring a shell means remaking the security case. The title still says "temporary"; the branch says "decided".
- **Body**: "The fail-closed sandbox settings (`failIfUnavailable`, `allowUnsandboxedCommands: false`, network deny-all) **stay**, inert … The test says so." — `77243fd` removed both the settings block and the test that pinned it. The body asserts the opposite of the diff.
- **Body**: "the mechanism and the one-line re-test probe live in the ADR amendment" — the amendment carries no re-test probe; the ponytail rounds trimmed it.
- **Body**: "95 tests pass" — 93 now, after the removal.

Rewrite title and body against `63c760c`.

### 2. The ADR amendment states as fact what only PR #691 delivers

> Everything that existed to serve the shell goes with it — no bubblewrap or socat in the image, no bwrap smoke check, and no `sandbox` settings in the SDK options or the managed-settings policy tier

Only the SDK-options clause is true at this commit. At `63c760c`: `infra/microvm-image/Dockerfile:19` still installs `bubblewrap socat`; `infra/microvm-image/smoke.sh:21-31` still runs the bwrap check; `infra/microvm-image/managed-settings.json:15-19` still carries the `sandbox` block. (All four are in #691's file list, so the sweep is covered — this is an *ordering* problem, not a missing-file problem.)

Two consequences:

- **Merge #691 first, or merge them together.** If this lands alone, ADR-0034 asserts three things that are false in the tree it describes.
- **Worth a moment's thought, not just a doc edit**: `settingSources: []` still reads the managed tier, and until #691 lands that tier says `sandbox.enabled: true, failIfUnavailable: true`. With Bash denied the sandbox should never be reached, so this is very likely inert — but the PR's own security case is "the sandbox path is gone", and on this branch alone it isn't. Either confirm the CLI never initializes the sandbox when every shell tool is denied, or land #691 with this.

Simplest fix: state that clause as what #691 lands, or gate the merge on #691.

### 3. The amendment's `#646` reference now dangles

> The #646 probe **this ADR relied on** only exercised namespace creation, never the proc mount.

But the diff deleted the sentence recording that reliance from *Measured facts this decision rests on*:

```diff
-Unprivileged user namespaces and bubblewrap work in the guest kernel at default
-capabilities (no `--additional-os-capabilities`); …
```

The ADR no longer says it relied on it, so the amendment corrects a claim the document no longer makes. That section is a record of what was measured and believed — history, not current state — and issue #692's own analysis makes this the crux ("how 'bubblewrap works at DEFAULT capabilities' entered the spec and ADR-0034 as settled fact").

This is also the repo's documented ADR practice: amendments correct, they don't rewrite. ADR-0030 spells it out — "Their records remain unchanged as the history of the tool surface they introduced" and "The Fargate-tool wording below records the original decision context." ADRs 0001, 0005, 0011, 0022, 0025, 0026, 0027 all follow it.

Restore the measured fact (marked as retracted by the amendment, or left plain with the amendment doing the correcting) and let the amendment be the single place the record is fixed. The status-line anchor is correct and the trailing-section format is fine — it's the body edit that departs from precedent.

---

## Non-blocking

- `turn-serving.test.ts:162,189` and `turn-stream-mapper.test.ts:117-170` still use `toolName: "Bash"` / `type: "tool-Bash"` as stream-mapper fixtures — pinning a shape the runtime can no longer produce. Harmless; any other tool name reads truer next time those are touched.
- `query-options.ts:27` keeps `SHELL` in `CLI_ENV_ALLOWLIST` (pinned at `query-options.test.ts:24,51`). Arguably shell-serving leftover — but the CLI may read it independently of the Bash tool and it is not a secret, so I would leave it rather than guess.
- `docs/agents/security.md:30` links the ADR file, not the `#amendment-2026-09-01--no-shell-in-the-vm` anchor the status line uses.
- ADR "Considered options": "without amputating the toolset **(the shell aside)**" quietly concedes the option's own premise in a parenthetical. A clause would carry it better.
- `CONTEXT.md:147-150` (Workspace) now wraps raggedly against the file's fill width.
- Pre-existing, flagging not asking: `doc-tools.ts:220` passes a parameter literally named `sandbox` — the exact word `CONTEXT.md` lists under *Avoid*. It comes from the shared `@mymemo/document-tools` signature, so it is out of scope here.

---

**Summary** — Standards: 3 documented-convention findings (ADR body rewritten rather than amended; two present-tense claims that need #691), 5 judgement calls. Spec: the decision is coherent, complete on this PR's side, and correctly implemented; the gap is that the record asserts an end-state the tree does not yet have. Nothing is wrong with the code. Fix the title/body, sequence #691, and restore the #646 measured fact, and this is ready.

CHANGES REQUIRED

chengchao:
All three blocking findings addressed on `f24ea7c`:

1. **Title and body rewritten.** They described an earlier draft — the body still claimed the fail-closed `sandbox` settings "stay, inert" after `77243fd` deleted them, the title still said "until the sandbox works" after the decision was reframed as settled, and it cited a re-test probe the amendment no longer contains plus a stale test count. Since the repo squash-merges, that text lands as the commit message, so it now matches the branch: settings removed, 93 tests, settled decision.

2. **The #691 dependency is named, not assumed.** The amendment no longer asserts image-side removals this branch doesn't make. It states that #691 must land with or before this change, and spells out the consequence you identified: until it does, `settingSources: []` still reads a managed tier whose `sandbox` block is live — inert with Bash denied, but not something this PR's security case should claim away. Merge order is now the first line of the PR body.

3. **The corrected measured fact is restored.** You're right that this departed from documented practice (ADR-0030, followed by 0001/0005/0011/0022/0025/0026/0027): an amendment corrects the record rather than rewrites it. Earlier ponytail rounds trimmed the false bubblewrap fact out entirely, which also left the amendment correcting a claim the document no longer made. The fact is back in "Measured facts", marked **corrected: see the amendment**, and the amendment now refers to it directly.

Also took the `security.md` anchor-link nit.

Left alone deliberately, as non-blocking: the `"Bash"` fixture names in the stream-mapper tests (renaming adds diff to exercise a mapper that is tool-agnostic), and `SHELL` in `CLI_ENV_ALLOWLIST` (it is the CLI's own environment, not a shell tool's, and dropping it is a behaviour change worth its own justification rather than a drive-by).

93 tests, tsc, biome clean; CI green on the prior head and re-running.

chengchao:
## Fresh review — `feat/drop-bash-tool` @ `f24ea7c` vs `origin/main` (`5561dcc`)

Reviewed the whole diff afresh (10 files, +67/−60) on the two axes the repo's
`/code-review` uses. Verified locally from the worktree: `bun test` in
`apps/in-vm-server` → **93 pass / 0 fail**, `bunx tsc --noEmit` → **clean**,
`bunx biome check apps/in-vm-server/src docs` → **clean**. Matches the body's
verification claims.

### The three findings from the previous round — all genuinely fixed

1. **Title/body vs. the diff.** The title (`feat(in-vm-server): no shell in the
   VM — deny Bash and remove the sandbox machinery`) and body now describe this
   head, not a draft. I walked every claim in the "What changes" section against
   the diff: the `disallowedTools` move, the removal of the `sandbox` block *and*
   its dedicated test, the ADR amendment, the `security.md` / `CONTEXT.md`
   follow-through, and the sweep across code comments, `.env.example`, and
   `docs/agents/configuration.md` — all present, nothing claimed that isn't
   there, nothing in the diff the body omits. As a squash-merge message it lands
   accurate.

2. **The ADR no longer asserts #691's content as done.** The amendment now reads
   "the image drops bubblewrap, socat, the bwrap smoke check, and the policy
   tier's own `sandbox` block (PR #691 …)". I pulled `gh pr diff 691` rather than
   trusting the reference: its head removes `bubblewrap socat` from the
   `dnf install` line, deletes the `sandbox` block from
   `infra/microvm-image/managed-settings.json`, deletes the `bwrap` stanza and
   `SKIP_BWRAP` from `smoke.sh`, and drops the `socat` assertion from
   `scripts/smoke/microvm-image-check.sh`. Every item the ADR attributes to #691
   is actually in #691. The interim consequence it states — "the image still
   configures a sandbox for a shell that cannot be invoked" — is exactly true of
   `managed-settings.json` on `main` today.

3. **The #646 measured fact is corrected in place, not deleted.** The claim is
   restored with `— **corrected: see the amendment**` bound to the bubblewrap
   clause by an em dash and closed by a semicolon before the four facts that
   still stand, and the amendment names the precise gap ("the #646 probe … only
   exercised namespace creation, never the proc mount"). No dangling reference:
   the section heading still reads "(probe #646, egress probe #651)" and now has
   a fact to point at. This matches the repo's ADR practice (cf. ADR-0030's
   "Amended (2026-08-22) by …", ADR-0033's status-line amendment chain).

### What I checked beyond that

- **The denials are real, not decorative.** `disallowedTools` entries are parsed
  through the SDK's rule parser (`c7` → `ud` in
  `claude-agent-sdk/sdk.mjs`), which normalizes legacy tool names through
  `{Task:"Agent", KillShell:"TaskStop", KillBash:"TaskStop",
  BashOutput:"TaskOutput", …}`. So `Bash` denies `Bash`, and `BashOutput` /
  `KillShell` normalize to the canonical `TaskOutput` / `TaskStop` — the names
  are valid and the rules bind. (Side effect, harmless: they also cover the
  Agent tool's output/stop, and `Agent` isn't in `allowedTools` either.)
- **Belt and braces on Bash.** It is both absent from `allowedTools` and present
  in `disallowedTools`, under `permissionMode: "dontAsk"`. Removing it from the
  allowlist alone would already have denied it; the explicit deny is what makes
  it testable, which is the stated point.
- **The sweep is complete on this side of the boundary.** Grepping
  `bubblewrap|bwrap|socat|sandboxed Bash|OS-sandboxed` across the repo leaves
  hits only in `infra/microvm-image/**` and `scripts/smoke/microvm-image-check.sh`
  (#691's files) plus retired-path ADRs and research notes, which correctly
  record history. `docs/agents/architecture.md:67` and `agentcore-runtime.md:34`
  are unaffected (generic bundle pointer; the separate Run path).
- **The anchor works.** `#amendment-2026-09-01--no-shell-in-the-vm` is the
  GitHub slug for `## Amendment 2026-09-01 — no shell in the VM` (em dash
  dropped, double space → double hyphen). Both the status line and the
  `security.md` link resolve.
- **The removed test wasn't load-bearing.** It only asserted the shape of the
  `sandbox` object that no longer exists. Coverage of what remains is a `toEqual`
  on both arrays, so a regression that reintroduces `Bash` or drops a denial
  fails. No tautology introduced (`CODING_STANDARDS.md`).
- **`doc-tools.ts`'s symlink guard.** Its comment's premise narrows now that the
  shell is gone (`Edit`/`Write` cannot themselves plant a symlink), but the guard
  is cheap, still correct, and the right thing to keep. No change wanted.

### Non-blocking

1. **Merge order: land them together, and never #691 alone first.** The ADR says
   #691 "must land with or before this change". Half of that permitted space is
   unsafe: if #691 lands *alone* on `main`, the image loses bubblewrap and the
   policy tier's `sandbox` block while `query-options.ts` on `main` still passes
   `sandbox: { enabled: true, failIfUnavailable: true }` — a fail-closed sandbox
   with no `bwrap` to build, whose startup behaviour #691's own body admits is
   untested. #691's body meanwhile assumes the opposite sequence ("With the shell
   now denied (#692/#693) …"). The safe orders are *together* (the "with" this
   sentence already offers first) or **this PR first**, whose interim state the
   ADR itself describes as benign. Suggest tightening the parenthetical to say so
   — e.g. "PR #691, which must land with this change or after it, never before" —
   so the durable record doesn't prescribe the one ordering that breaks `main`.
2. `SHELL` still sits in `CLI_ENV_ALLOWLIST` (`query-options.ts:27`). Harmless
   and arguably still wanted by the CLI itself, but it is a leftover of the era
   this PR ends; worth a look next time that list is touched.
3. Two ragged rewraps left by the sweep — `CONTEXT.md`'s Workspace entry
   ("replacement via the Checkpoint. Not the Agent session, and not" /
   "durable beyond the Checkpoint.") and `doc-tools.ts:90-91`. Biome doesn't
   reflow prose, so these are cosmetic only.
4. `feat(` for a change that is net removal plus docs is a stretch; `refactor(`
   or `fix(` would read truer in the log. Not worth a round-trip on its own.

Nothing in the diff is wrong, incoherent with ADR-0034, or in breach of a
documented standard. The security argument for denying rather than weakening
(IMDS → the `conversations/*` checkpoint scope whose residual the ADR accepted
*because* the process boundary contained it, plus gateway spend and DNS
exfiltration) is the right call and is recorded where it will be found.

APPROVED FOR MERGE

chengchao:
## Fresh review — `feat/drop-bash-tool` @ `fc28c0c` vs `origin/main` (`5561dcc`)

Reviewed the full diff independently on the repo's two `/code-review` axes
(10 files, +68/−60). Verified from the worktree: `bun test apps/in-vm-server/`
→ **93 pass / 0 fail**, `bunx tsc --noEmit -p apps/in-vm-server` → **clean**,
`bunx biome check` → **clean**. All five CI checks now green, `image-contract`
included.

### The head's one commit — the merge-order correction — is right, and understates its own case

`git diff f24ea7c..fc28c0c` is exactly the one ADR hunk; nothing else moved.
Reasoning through all three orderings:

- **#693 alone first** — Bash denied in the SDK options; the image still ships
  bubblewrap and a policy-tier `sandbox` block that nothing consults. Dead
  config, no behaviour change. Safe.
- **#691 alone first** — `main` still passes `sandbox: { enabled: true,
  failIfUnavailable: true }` with no `bwrap` to build it. The SDK's own typings
  (`sdk.d.ts`) say that when sandbox dependencies are missing, `query()` "will
  emit an error result and exit rather than silently running commands
  unsandboxed" — so this takes out **every Turn**, not just Bash calls. The ADR
  now describes it more mildly ("the tool demanding a sandbox that cannot be
  built"); the real blast radius is larger. Correctly forbidden either way.
- **Together** — clean.

So "landing with this change or after it — never before" is the correct clause.

### The three round-1 findings — genuinely fixed

1. **Title/body vs. the diff** — title now describes this head; every "What
   changes" claim walks against the diff (see one regression below).
2. **The ADR no longer asserts #691's content as done** — it attributes the
   image-side removals to #691 rather than claiming them.
3. **The `#646` reference is no longer dangling** — the measured fact is
   **marked corrected in place** ("**corrected: see the amendment**"), keeping
   the probe attribution instead of deleting the sentence.

---

## Blocking

**1. The PR body still prescribes the ordering this head's commit exists to repudiate.**

The ADR now reads:

> (PR #691, landing with this change or after it — **never before** …)

The PR body still reads, verbatim:

> **Merge order: land #691 with or before this PR.** … Until it lands, the
> image still configures a sandbox for a shell that can no longer be invoked,
> **and the ADR amendment here names that dependency.**

Two problems. It instructs whoever merges to run the one sequence that breaks
`main` — and the merger reads the PR body, not the ADR buried in the diff. And
its closing clause now misdescribes the ADR it points at: the amendment names
the *opposite* dependency direction. This body is also the squash-merge message,
so the repudiated ordering would be written into `main`'s permanent history in
the very commit that corrects it.

Notably **#691's body has already been fixed** ("land this with #693, or after
it — never before"), so #693's is the only one left carrying the old text —
this reads as a missed spot, not a disagreement. One-line fix, no code change.

---

## Non-blocking

- **The amendment credits the wrong mechanism.** "`Bash`, `BashOutput`, and
  `KillShell` therefore sit in `disallowedTools` … denied, not merely unlisted"
  — accurate, but `disallowedTools` is defence-in-depth here, not the closer.
  What actually closes the shell is `permissionMode: "dontAsk"` plus the
  exact-equality allowlist: the SDK's tool union also carries `REPL`
  ("JavaScript code to execute. Supports top-level await. State persists across
  calls."), `Workflow`, and `Agent`, none of which are denied and all of which
  are closed by that same allowlist. The confinement is sound as built — the
  test pins `allowedTools` by equality — but a reader who takes the amendment
  literally may conclude enumeration in `disallowedTools` is what confines, and
  add a tool to `allowedTools` believing the denials still cover them.
- **Removing the `sandbox` block was genuinely safe**, verified rather than
  assumed: `strictAllowlist` is "enforced for sandboxed commands only — in-process
  tools such as WebFetch are not gated by this setting", and the removed block
  carried no `filesystem`/`credentials` keys. It constrained nothing but Bash.
- **Stale fixtures.** `turn-stream-mapper.test.ts:119,137,160` and
  `turn-serving.test.ts:162,189` still use `toolName: "Bash"` / `type: "tool-Bash"`
  for a tool the VM can never emit. The mappers are tool-agnostic so nothing is
  wrong mechanically, but it reads as evidence Bash still flows.
- **`SHELL` remains in `CLI_ENV_ALLOWLIST`** (`query-options.ts:27`) — inert with
  no shell to spawn, carried over from the previous round.
- **`app.ts`'s `smokeScriptPath` JSDoc** now points at a smoke script that, on
  this branch, still reports bwrap. Self-corrects when #691 lands.
- **Pre-existing, out of scope:** `doc-tools.ts:220` passes a parameter literally
  named `sandbox` (from `@mymemo/document-tools`, a Vercel-Sandbox-era name) —
  against CONTEXT.md's `_Avoid_: sandbox`. And `query-options.ts:11` cites
  `research/agent-sdk-confinement`, which exists nowhere in the repo; dangling
  before this PR.
- **Doc sweep is complete in scope.** Every surviving `bwrap`/`bubblewrap`/
  `failIfUnavailable`/"sandboxed Bash" string is in #691's files
  (`infra/microvm-image/`, `scripts/smoke/`), the ADR's own deliberate history,
  or a dated research note citing Anthropic's public docs. The ADR anchor
  `#amendment-2026-09-01--no-shell-in-the-vm` is correct — the em dash is
  deleted by GitHub's slugger while both flanking spaces survive as hyphens, so
  the double hyphen is right, and both links use it.

### Axis summary

**Standards:** no documented-standard violations; judgement calls only, none above
noise. The change is subtractive and the rationale now lives once. **Spec:** no
scope creep, no missing intent, denial correctly implemented — one blocking
inconsistency between the merged record and the PR body.

Fix the merge-order line in the body and this is ready.

CHANGES REQUIRED

chengchao:
Both addressed on `a4a040f`.

**Blocking — the PR body's merge order.** Fixed, and thank you for catching it: I corrected the ADR and #691's body but missed this one, which is the copy that actually becomes the squash-merge message. It now reads "land #691 with this PR or after it — never before," states why that direction is the unsafe one, and notes that the reverse gap is harmless. Your sharper reading is in there too — a missing sandbox dependency under `failIfUnavailable: true` takes down the whole `query()`, not just Bash calls, so #691-first would kill every Turn rather than degrade one tool.

**Non-blocking — the mechanism framing.** Fixed rather than deferred, because being wrong about which control enforces a security boundary is how the next person makes a mistake. The amendment now says plainly that `permissionMode: "dontAsk"` against an exact allowlist is what closes the shell and everything else the SDK ships, that a tool added to `allowedTools` is admitted regardless of the deny list, and that the denials are defence in depth plus a statement of intent. The code comment says the same in two lines.

Left as-is, as you marked them: the `"Bash"` stream-mapper fixtures and `SHELL` in `CLI_ENV_ALLOWLIST`.

93 tests, tsc, biome clean.

chengchao:
## Independent review — `feat/drop-bash-tool` @ 3b943c3 vs `origin/main` (5561dcc)

Fresh two-axis review (Standards + Spec) plus my own verification of the fixes the earlier rounds claimed. 10 files, 14 commits.

### Verified green

- `bun test apps/in-vm-server` — 93 pass / 0 fail. `bunx tsc --noEmit -p apps/in-vm-server` clean. `bunx biome check` on the five changed TS files clean.
- `docs/agents/security.md` matches `query-options.ts` key for key (`settingSources: []`, `strictMcpConfig`, `permissionMode: "dontAsk"`, `Read(./**)`/`Edit(./**)`, the three deny entries), and `query-options.test.ts` pins all of it against literals — no tautology (CODING_STANDARDS.md).
- Anchors resolve: `## Amendment 2026-09-01 — no shell in the VM` → `#amendment-2026-09-01--no-shell-in-the-vm`, used correctly by both the ADR status line and security.md's `../adr/...` link.
- Stale-reference sweep is complete. Repo-wide grep for `sandbox|bwrap|bubblewrap|socat|failIfUnavailable|allowUnsandboxedCommands` leaves zero residue in `apps/in-vm-server/**` and `CONTEXT.md`. Remaining hits are all legitimate: `infra/microvm-image/**` + `scripts/smoke/microvm-image-check.sh` (#691's scope), the still-live AgentCore/Harness paths in `docs/agents/agentcore-runtime.md`, `security.md:24`, `chat-api.md`, and historical ADRs 0004/0006/0030/0033.
- The "measured facts" correction is factually right: `infra/microvm-image/smoke.sh:28` runs `bwrap --unshare-all --ro-bind / / --dev /dev true` — no `--proc`, corroborating "only exercised namespace creation, never the proc mount."

### The four earlier-round fixes all hold on this head

1. **Merge order** — safe in both places. ADR: "PR #691, landing with this change or after it — never before". PR body: "**Merge order: land #691 with this PR or after it — never before.**" Rationale is sound in both directions.
2. **#691 stated as pending, not done** — the ADR says "landing with this change or after it", not asserting it as complete.
3. **#646 reference** — no longer dangling; it now names what the probe did and didn't exercise.
4. **Enforcement attribution** — the amendment now reads "What closes the shell ... is `permissionMode: \"dontAsk\"` against an exact allowlist ... The denials are defence in depth", and the overstated precedence claim is gone.

### On (4) specifically — I checked the SDK, and the amendment is right

Both of my axis reviews independently argued the correction over-swung, citing `sdk.d.ts`: `disallowedTools` is "removed from the model's context and cannot be used, **even if they would otherwise be allowed**", whereas `dontAsk` only "denies if not **pre-approved**" — and since `settingSources: []` still reads the managed policy tier, that tier looked like a second pre-approval source that only `disallowedTools` could beat.

That argument does not survive checking the tier's semantics. `sdk.d.ts:2925-2929` — `managedSettings` "is filtered through a **restrictive-key allowlist** (`allowManaged*Only` locks, `permissions.deny`/`ask`, sandbox restrictions); non-restrictive keys ... are silently dropped." The policy tier can carry `deny`/`ask`, never `allow`, and `infra/microvm-image/managed-settings.json` carries only `deny` + locks. So there is no allow-source anywhere in this configuration, "anything not named in the allowlist is denied" is exactly true, and describing the deny entries as defence in depth is accurate rather than an understatement. The claim the last commit removed reached a true conclusion by the wrong route; removing it was correct. **No change needed here** — recording this so a fifth round doesn't reverse it again.

### Non-blocking observations

- **PR body still carries the retracted claim.** "With `permissionMode: \"dontAsk\"` deny beats allow, so the managed policy tier cannot reopen it." The merge-order paragraph was corrected but this sentence was not, so the body and the ADR now disagree on the same point, and the attribution is muddled either way (deny-beats-allow is permission-rule precedence, not a property of `permissionMode`). Worth a one-line edit before squash, since the body lands as the commit message.
- **ADR-0034 edits its original decision prose in place** (summary, trust-boundary bullet, considered-options bullet), where ADR-0033's precedent is append-only with a pointer ("Stage 2 is decided in the amendment below."). No documented rule either way in `docs/agents/domain.md`, the inline `**corrected: see the amendment**` marker on the measured facts is the right pattern, and leaving "confined by OS-sandboxed Bash" atop the governing ADR would be its own falsehood. Judgement call, and the PR body shows it was a considered one.
- `it("scopes the file tools to the cwd and denies the shell outright")` also asserts `WebFetch`/`WebSearch`, which are not the shell. Pre-existing overloading; rename or split if you touch it.
- `SHELL` remains in `CLI_ENV_ALLOWLIST` (`query-options.ts:27`) — now unused by any tool. Harmless; not worth a commit on its own.
- The amendment has one ragged wrap ("The denials are defence in depth. Running the shell unsandboxed was / rejected:") in an otherwise hard-wrapped document.

### Summary

No defects in the merged content. The security reasoning is sound and, on the one point three rounds have circled, verifiably correct as written. The only thing I would touch before merging is the PR body sentence above, which is a description edit, not a code change.

APPROVED FOR MERGE

chengchao:
Body corrected — it still carried the retracted "deny beats allow, so the managed policy tier cannot reopen it" line, which would have landed as the squash message contradicting the ADR. It now matches the amendment: the exact allowlist under `permissionMode: "dontAsk"` is the enforcement, the deny entries are defence in depth. Body-only, so the head is unchanged at `3b943c3`.

Thank you for checking the `managedSettings` filtering rather than accepting your sub-agents' argument at face value — that the restrictive-key allowlist admits `permissions.deny`/`ask` but never `allow`, so no allow-source exists in this configuration. That settles the question in the right direction and is worth having on the record so a later round doesn't reverse it.

The three remaining nits are noted and deliberately not taken here: the overloaded test name, `SHELL` in `CLI_ENV_ALLOWLIST` (it is the CLI's own environment rather than a shell tool's, so removing it is a behaviour change deserving its own justification), and the in-place ADR edit versus ADR-0033's append style — the latter was itself a fix demanded by an earlier round citing ADR-0030, so it stays as-is rather than oscillating between conventions.

Issue #658, opened 2026-08-31
v2: UIMessage Live Stream lane on the Redis relay


chengchao:
## Parent

Spec #654 — /v2 chat on Lambda MicroVMs.

## What to build

A v2 lane in the relay package (live-text) that carries a Turn's UIMessage events over the existing payload-agnostic Redis transport: a publisher side for the In-VM server and a subscriber side for chat-api. Per-Turn channels; Assistant text deltas exist only here; terminal detection from the Turn's terminal chunk. Deliberately NOT ported from the v1 lane: the producer-buffered backlog and request/reply re-attach — the v2 Live Stream dies with its Turn, and a late or disconnected subscriber recovers from durable history instead.

## Chunk-schema contract (amended 2026-08-31 — one UIMessage per Turn, stock grammar)

The lane's payload grammar is the **stock AI SDK v7 UI Message Stream chunk vocabulary** — https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol — validated against `ai@7.0.83`'s `UIMessageChunk` types (already resolved in the lockfile). **No custom chunk types.**

**Framing: one assistant UIMessage per Turn.** Verified in the `ai@7.0.83` source: `useChat`'s `processUIMessageStream` holds exactly one assistant message per stream — a second `start` chunk mutates the current message's id rather than opening a new message — and AI SDK's own agents (`ToolLoopAgent`, `WorkflowAgent`) stream an entire multi-step loop as one UIMessage. Provider-call boundaries are `start-step`/`finish-step` chunks inside the message; tool parts are embedded.

**Terminal chunks ARE the Turn's Outcome** (possible because the Turn is one message): `finish` (done) / `abort` (interrupted) / `error` (error). Turn status additionally rides `finish.messageMetadata` (v7 carries `messageMetadata` on `start`/`finish` plus a `message-metadata` chunk — source-verified). There is no custom Outcome event; `data-*` parts stay free for the generative-UI follow-on.

Layering split (jointly "the standard UIMessage SSE" of spec #654): the lane carries the **JSON chunk payloads only**. The SSE envelope — `x-vercel-ai-ui-message-stream: v1` header, `data:` framing, `[DONE]` terminator, keepalive pings — is the HTTP response's job in #667, not this lane's.

## Acceptance criteria

- [x] Producer publishes validated UIMessage chunks on a per-Turn channel; size caps and oversize behaviour defined and tested.
[...]

PR #694, opened 2026-09-01
feat(chat-api): v2 message POST + own-Turn UIMessage SSE (#667)

### LINKED ISSUE #667: v2: message POST + own-Turn SSE

chengchao:
Closes #667. Parent: spec #654 (/v2 chat on Lambda MicroVMs).

## What this builds

**`POST /v2/conversations/:conversationId/messages`** — submit a UIMessage; the response is the stock AI SDK v7 UI Message Stream scoped to the submitted message's own Turn. This route owns the SSE envelope around the chunk payloads the #658 lane relays: `x-vercel-ai-ui-message-stream: v1` (and the rest of `UI_MESSAGE_STREAM_HEADERS`), `data:` framing, `: ping` keepalives, `data: [DONE]`.

Order of operations: identity → `503` while no In-VM server is configured → exposure gate (`403`, before any write) → the `queued` INSERT → **subscribe to the Turn's Live Stream lane** → **nudge** → relay.

- **Store** (`conversation-messages` feature, `enqueueTurn`/`getTurnStatus` added to the existing `ConversationMessagesStore`): `enqueueTurn` runs under the Conversation row lock (`SELECT … FOR UPDATE`, the same lock `PATCH` archive takes), so `404` (missing/foreign), `409` (archived), `queued`, and `duplicate` (with the existing Turn's status) come out of one atomic step and no message slips in beside an Archive. It calls agent-db's `enqueueTurnTx`, now typed to accept a `DbTx` too. This INSERT is chat-api's only write to `conversation_messages`; the In-VM server owns every status transition.
- **Route**: the body is the stock `useChat`/`DefaultChatTransport` request (`.strict()`: `id` = path Conversation id, `trigger: "submit-message"`, `messages`); only the final message is submitted and must be a user UIMessage with text parts, its client `id` becoming the Turn id (held to the path-safe id shape since it names the relay channel). Earlier messages are validation input only, never durable history (the v1 stance). No `409` for concurrency: a second POST is the next queued row; its stream holds with silent `: ping` keepalives while predecessors drain, then carries only its own Turn's chunks.
- **Live Stream lane re-keyed** (`packages/live-text`, from review): the Turn relay's channel was keyed on the client-chosen message id alone, which is unique only within its Conversation (the table's PK), so a reused id in another Conversation would have shared a channel. `openPublisher`/`subscribe` now take `{ conversationId, messageId }` (Conversation id as the Redis hash tag), pinned by a contract test on both transports; the In-VM publisher and this route pass the pair.
- **Terminal watch**: the 5 s keepalive tick also reads the Turn's status; two consecutive terminal reads with no terminal chunk received (publisher died, restarted VM swept it `interrupted` — a single read may only be the In-VM server's commit-before-publish window) end the stream with `[DONE]` and no synthesized chunk. A failed status read is logged and retried next tick (only a failed write ends the stream); a finished stream never re-arms its tick. A mid-stream relay failure closes the response without `[DONE]`; a subscribe failure answers `503` after a best-effort nudge; a nudge failure is logged and the response streams on (the row is durable, the In-VM server self-heals on its interval). In every case the client Recovers from durable history.
- **Duplicate client message id** (documented in `docs/agents/chat-api.md`): no second Turn — the PK is the idempotency authority and the existing row is untouched. Still `queued`/`processing` → the response attaches to its Live Stream and nudges (earlier chunks missed; no backlog). Already terminal → `410 { error: "Turn already ended", recovery: "history" }`, the v1 history-recovery signal. A client id that names an assistant message (ids are visible in history) → `409`, nothing written.
- **Client disconnect** drops chat-api's subscription and nothing else; the Turn runs to its Outcome.
- **Ensure-VM is the dev-mode stub**: `IN_VM_SERVER_URL` (new optional `ApiConfig.inVmServerUrl`) → `AppDeps.nudgeInVmServer` POSTs `new URL("/nudge", url)` with a 5 s timeout; unset → `503`. With `IN_VM_SERVER_AUTH_TOKEN` (the platform's per-VM token) the nudge also sends `X-aws-proxy-auth` + `X-aws-proxy-port: 8080`, so the route can be driven against a hand-launched MicroVM before orchestration mints tokens itself. The orchestration ticket replaces this seam with per-Conversation MicroVM launch. `AppDeps` also gains `turnLiveStreamRelay` (Redis Turn lane, closed with the v1 relay on shutdown).
- `ai@7.0.83` becomes a chat-api dependency (already in the lockfile via live-text) for `UI_MESSAGE_STREAM_HEADERS` and the client-side test parser.

## Acceptance criteria

- **Against a locally running In-VM server: POST streams the full Turn as UIMessage SSE, ends with the Outcome, and the durable history matches the streamed message — verified live** (Compose Postgres + Redis, `apps/in-vm-server` on OpenRouter, chat-api local composition with `IN_VM_SERVER_URL`): `200 text/event-stream` + `x-vercel-ai-ui-message-stream: v1`; frames `start` → `start-step` → `text-start` → `text-delta`×3 → `text-end` → `finish-step` → `finish {"messageMetadata":{"status":"done"}}` → `[DONE]`; `GET …/messages` then returned the user row `status: done` and the assistant row with the same `messageId` and `[{step-start},{text:"pelican",state:"done"}]`.
- **Consumable by a stock v7 `useChat`/`DefaultChatTransport` with no custom transport** — `conversation-messages.route.test.ts` drives the real `DefaultChatTransport` (`ai@7.0.83`) through the app and `readUIMessageStream`; the Turn renders as one assistant message with `[step-start, text]`.
- **Two rapid POSTs: no 409; the second holds with silent keepalives during its predecessor, then streams only its own Turn's events** — live (two parallel `curl`s: second stream = 2 × `: ping`, then only its own 8 frames, zero leakage from the predecessor's 41) and in a route test (fake drain loop, keepalive asserted).
- **Subscribe-before-nudge ordering tested — no lost early events** — route test whose fake In-VM server publishes the whole Turn *synchronously inside the nudge call*; every chunk arrives.
- **Duplicate client message id: no second Turn; behaviour documented and tested** — store test (pglite: second enqueue is `duplicate`, parts untouched), route test (`410` for a terminal Turn with no nudge; attach for an in-flight one), live (`410 {"error":"Turn already ended","recovery":"history"}`), docs section.
- **Client disconnect mid-stream leaves the Turn running to its Outcome (verified via durable history)** — live: `curl --max-time 2` gave up after 1 frame; history showed the Turn `done` ~8 s later with the full 21-line assistant message. Route test covers the same with an aborted request.
- **Archived Conversation refuses the message; the exposure gate is enforced on submission** — live `409 {"error":"Conversation is archived"}` on an archived Conversation and `404` for a foreign owner; route test for the gate (`403`, no write, no nudge). (The local composition's gate is always-open, so the gate is unit-tested only.)

## Notes for review

- **`409` for archived, none for concurrency.** The ticket's "no 409 anywhere" is about the second POST; "archived-blocks-new-messages" is carried from the v1 message path where it is `409`, so that one stays.
- **Body shape.** "Submit a UIMessage" is read as the stock `useChat` body, which is what "consumable by a stock `DefaultChatTransport`" needs without `prepareSendMessagesRequest`. `regenerate-message`, file parts, and message metadata are `400`.
- The exposure gate runs on every submission, duplicates included (simplest rule; v1 skipped it on exact retries).
- The keepalive interval stays the v1 constant (5 s); the two timing-dependent tests wait for one tick, as the v1 keepalive test does.
- Pre-existing, untouched: chat-api's `bunx tsc --noEmit` reports the known `s3-artifact-download-signer.ts` `@smithy/types` fork error present on `main`.

## Verification

- `apps/chat-api`: 190 pass (25 in the conversation-messages feature: 15 route, 10 store); `bunx tsc --noEmit` clean apart from the pre-existing signer error.
- `packages/agent-db`: 254 pass, 17 skip; `bunx tsc --noEmit` clean.
- `packages/live-text`: 53 pass (new contract test: same message id, two Conversations, two streams); `bunx tsc --noEmit` clean.
- `apps/in-vm-server`: 93 pass; `bunx tsc --noEmit` clean.
- `biome check` clean on the touched workspaces.
- Live end-to-end as above; the Compose stack was torn down with its volume and the key file shredded afterwards.
- **Real topology (see the "Real-topology verification" comment):** a scratch MicroVM image built from this branch, one hand-launched VM through the prod egress connector, and a one-off Fargate task running this branch's chat-api against prod RDS + Valkey, nudging the VM through its JWE endpoint and calling the model through the prod gateway — **14/14 checks pass** (every AC above plus a 75 s idle-subscriber probe on Valkey). All resources torn down.

https://claude.ai/code/session_01X7wZP25CP94qTbCfws5WYA

### DIFF
--- a/apps/chat-api/package.json
+++ b/apps/chat-api/package.json
@@ -20,6 +20,7 @@
 		"@mymemo/document-tools": "workspace:*",
 		"@mymemo/live-text": "workspace:*",
 		"@statsig/statsig-node-core": "^0.19.8",
+		"ai": "7.0.83",
 		"drizzle-orm": "^0.45.2",
 		"hono": "^4.10.1",
 		"hono-pino": "^0.10.3",
--- a/apps/chat-api/src/config/env.test.ts
+++ b/apps/chat-api/src/config/env.test.ts
@@ -83,9 +83,10 @@ describe("loadApiConfigFromEnv — ADR-0034 gateway credential rule", () => {
 		expect(loadApiConfigFromEnv(baseEnv()).inVmServerUrl).toBeUndefined();
 		const env = baseEnv();
 		env.IN_VM_SERVER_URL = "http://127.0.0.1:8080";
-		expect(loadApiConfigFromEnv(env).inVmServerUrl).toBe(
-			"http://127.0.0.1:8080",
-		);
+		env.IN_VM_SERVER_AUTH_TOKEN = "per-vm-token";
+		const config = loadApiConfigFromEnv(env);
+		expect(config.inVmServerUrl).toBe("http://127.0.0.1:8080");
+		expect(config.inVmServerAuthToken).toBe("per-vm-token");
 	});
 
 	it("still never reads the remaining worker-only secrets (KB, E2B)", () => {
--- a/apps/chat-api/src/config/env.test.ts
+++ b/apps/chat-api/src/config/env.test.ts
@@ -79,10 +79,10 @@ describe("loadApiConfigFromEnv — ADR-0034 gateway credential rule", () => {
 		expect(config.gatewayTokenSecret).toBe("gateway-signing-secret");
 	});
 
-	it("reads the dev-mode In-VM server URL for the v2 message POST, trailing slash dropped", () => {
+	it("reads the dev-mode In-VM server URL for the v2 message POST", () => {
 		expect(loadApiConfigFromEnv(baseEnv()).inVmServerUrl).toBeUndefined();
 		const env = baseEnv();
-		env.IN_VM_SERVER_URL = "http://127.0.0.1:8080/";
+		env.IN_VM_SERVER_URL = "http://127.0.0.1:8080";
 		expect(loadApiConfigFromEnv(env).inVmServerUrl).toBe(
 			"http://127.0.0.1:8080",
 		);
--- a/apps/chat-api/src/config/env.test.ts
+++ b/apps/chat-api/src/config/env.test.ts
@@ -79,6 +79,15 @@ describe("loadApiConfigFromEnv — ADR-0034 gateway credential rule", () => {
 		expect(config.gatewayTokenSecret).toBe("gateway-signing-secret");
 	});
 
+	it("reads the dev-mode In-VM server URL for the v2 message POST, trailing slash dropped", () => {
+		expect(loadApiConfigFromEnv(baseEnv()).inVmServerUrl).toBeUndefined();
+		const env = baseEnv();
+		env.IN_VM_SERVER_URL = "http://127.0.0.1:8080/";
+		expect(loadApiConfigFromEnv(env).inVmServerUrl).toBe(
+			"http://127.0.0.1:8080",
+		);
+	});
+
 	it("still never reads the remaining worker-only secrets (KB, E2B)", () => {
 		const env = baseEnv();
 		env.KB_DATABASE_URL = "postgresql://kb:kb@localhost:5432/mymemo_kb";
--- a/apps/chat-api/src/config/env.ts
+++ b/apps/chat-api/src/config/env.ts
@@ -120,7 +120,6 @@ export function loadApiConfigFromEnv(env: Env): ApiConfig {
 		openrouterBaseUrl:
 			env.OPENROUTER_BASE_URL?.trim() || "https://openrouter.ai/api",
 		gatewayTokenSecret: env.GATEWAY_TOKEN_SECRET?.trim() || undefined,
-		inVmServerUrl:
-			env.IN_VM_SERVER_URL?.trim().replace(/\/+$/, "") || undefined,
+		inVmServerUrl: env.IN_VM_SERVER_URL?.trim() || undefined,
 	};
 }
--- a/apps/chat-api/src/config/env.ts
+++ b/apps/chat-api/src/config/env.ts
@@ -69,6 +69,13 @@ export interface ApiConfig {
 	 * never leaves this process. Optional alongside `openrouterApiKey`.
 	 */
 	gatewayTokenSecret?: string;
+	/**
+	 * Dev-mode Ensure-VM stub (ticket #667): the base URL of one locally
+	 * running In-VM server that `POST /v2/conversations/:id/messages` nudges.
+	 * The orchestration ticket replaces it with per-Conversation MicroVM
+	 * launch. While unset the v2 message POST answers 503.
+	 */
+	inVmServerUrl?: string;
 }
 
 /**
@@ -113,5 +120,7 @@ export function loadApiConfigFromEnv(env: Env): ApiConfig {
 		openrouterBaseUrl:
 			env.OPENROUTER_BASE_URL?.trim() || "https://openrouter.ai/api",
 		gatewayTokenSecret: env.GATEWAY_TOKEN_SECRET?.trim() || undefined,
+		inVmServerUrl:
+			env.IN_VM_SERVER_URL?.trim().replace(/\/+$/, "") || undefined,
 	};
 }
--- a/apps/chat-api/src/config/env.ts
+++ b/apps/chat-api/src/config/env.ts
@@ -76,6 +76,12 @@ export interface ApiConfig {
 	 * launch. While unset the v2 message POST answers 503.
 	 */
 	inVmServerUrl?: string;
+	/**
+	 * With `inVmServerUrl` pointing at a real MicroVM endpoint: the platform's
+	 * per-VM auth token (`CreateMicrovmAuthToken`), sent as `X-aws-proxy-auth`
+	 * on the nudge. Unset for a plain local In-VM server.
+	 */
+	inVmServerAuthToken?: string;
 }
 
 /**
@@ -121,5 +127,6 @@ export function loadApiConfigFromEnv(env: Env): ApiConfig {
 			env.OPENROUTER_BASE_URL?.trim() || "https://openrouter.ai/api",
 		gatewayTokenSecret: env.GATEWAY_TOKEN_SECRET?.trim() || undefined,
 		inVmServerUrl: env.IN_VM_SERVER_URL?.trim() || undefined,
+		inVmServerAuthToken: env.IN_VM_SERVER_AUTH_TOKEN?.trim() || undefined,
 	};
 }
--- a/apps/chat-api/src/deps.ts
+++ b/apps/chat-api/src/deps.ts
@@ -24,10 +24,7 @@ import { PostgresArtifactMetadataStore } from "./features/artifacts/postgres-art
 import { createS3ArtifactDownloadSigner } from "./features/artifacts/s3-artifact-download-signer";
 import type { ConversationHistoryStore } from "./features/conversation-history/conversation-history-store";
 import { PostgresConversationHistoryStore } from "./features/conversation-history/postgres-conversation-history-store";
-import type {
-	ConversationMessagesStore,
-	TurnSubmissionStore,
-} from "./features/conversation-messages/conversation-messages-store";
+import type { ConversationMessagesStore } from "./features/conversation-messages/conversation-messages-store";
 import { PostgresConversationMessagesStore } from "./features/conversation-messages/postgres-conversation-messages-store";
 import type { ConversationStore } from "./features/conversation-store/conversation-store";
 import { PostgresConversationStore } from "./features/conversation-store/postgres-conversation-store";
@@ -64,7 +61,7 @@ export interface AppDeps {
 	/** Permanent AG-UI Conversation-history projection over Postgres Runs. */
 	conversationHistoryStore: ConversationHistoryStore;
 	/** /v2 UIMessage history over `conversation_messages` plus the `queued` Turn INSERT — chat-api's only write; the In-VM server owns every status transition. */
-	conversationMessagesStore: ConversationMessagesStore & TurnSubmissionStore;
+	conversationMessagesStore: ConversationMessagesStore;
 	/** Durable split-runtime run queue and event log. */
 	runStore: RunStore;
 	/** Producer-buffered per-Run relay used by initial and reconnect SSE. */
@@ -151,9 +148,12 @@ export function createDeps(
 		// orchestration ticket swaps in per-Conversation MicroVM launch here.
 		nudgeInVmServer: config.inVmServerUrl
 			? async () => {
-					const response = await fetch(`${config.inVmServerUrl}/nudge`, {
-						method: "POST",
-					});
+					const response = await fetch(
+						new URL("/nudge", config.inVmServerUrl),
+						{
+							method: "POST",
+						},
+					);
 					if (!response.ok) {
 						throw new Error(`nudge answered ${response.status}`);
 					}
--- a/apps/chat-api/src/deps.ts
+++ b/apps/chat-api/src/deps.ts
@@ -6,8 +6,10 @@ import {
 import {
 	createLiveStreamTelemetry,
 	createRedisLiveStreamRelay,
+	createRedisTurnLiveStreamRelay,
 	type LiveStreamRelay,
 	type LiveStreamTelemetry,
+	type TurnLiveStreamRelay,
 } from "@mymemo/live-text";
 import type { Env as PinoEnv } from "hono-pino";
 import type { ApiConfig } from "./config/env";
@@ -22,7 +24,10 @@ import { PostgresArtifactMetadataStore } from "./features/artifacts/postgres-art
 import { createS3ArtifactDownloadSigner } from "./features/artifacts/s3-artifact-download-signer";
 import type { ConversationHistoryStore } from "./features/conversation-history/conversation-history-store";
 import { PostgresConversationHistoryStore } from "./features/conversation-history/postgres-conversation-history-store";
-import type { ConversationMessagesStore } from "./features/conversation-messages/conversation-messages-store";
+import type {
+	ConversationMessagesStore,
+	TurnSubmissionStore,
+} from "./features/conversation-messages/conversation-messages-store";
 import { PostgresConversationMessagesStore } from "./features/conversation-messages/postgres-conversation-messages-store";
 import type { ConversationStore } from "./features/conversation-store/conversation-store";
 import { PostgresConversationStore } from "./features/conversation-store/postgres-conversation-store";
@@ -58,12 +63,23 @@ export interface AppDeps {
 	conversationStore: ConversationStore;
 	/** Permanent AG-UI Conversation-history projection over Postgres Runs. */
 	conversationHistoryStore: ConversationHistoryStore;
-	/** /v2 UIMessage history over `conversation_messages` (read-only here; the In-VM server is the sole writer). */
-	conversationMessagesStore: ConversationMessagesStore;
+	/** /v2 UIMessage history over `conversation_messages` plus the `queued` Turn INSERT — chat-api's only write; the In-VM server owns every status transition. */
+	conversationMessagesStore: ConversationMessagesStore & TurnSubmissionStore;
 	/** Durable split-runtime run queue and event log. */
 	runStore: RunStore;
 	/** Producer-buffered per-Run relay used by initial and reconnect SSE. */
 	liveStreamRelay: LiveStreamRelay;
+	/** The v2 per-Turn UIMessage lane (#658) the message POST subscribes to. */
+	turnLiveStreamRelay: TurnLiveStreamRelay;
+	/**
+	 * Ensure-VM + nudge for one Conversation (#667). Undefined = not configured,
+	 * so the v2 message POST answers 503. Today a dev-mode stub over
+	 * `IN_VM_SERVER_URL`; the orchestration ticket replaces it.
+	 */
+	nudgeInVmServer?: (ref: {
+		userId: string;
+		conversationId: string;
+	}) => Promise<void>;
 	/** Cardinality-safe, payload-free Live Stream relay observability. */
 	liveStreamTelemetry: LiveStreamTelemetry;
 	/** Close the lazy Redis relay clients during service shutdown. */
@@ -117,6 +133,11 @@ export function createDeps(
 		deployment: "current",
 		telemetry: liveStreamTelemetry,
 	});
+	const turnLiveStreamRelay = createRedisTurnLiveStreamRelay({
+		url: config.redisUrl,
+		deployment: "current",
+	});
+	const inVmServerUrl = config.inVmServerUrl;
 	return {
 		config,
 		createHarnessChatAgent,
@@ -129,8 +150,23 @@ export function createDeps(
 		conversationMessagesStore: new PostgresConversationMessagesStore(database),
 		runStore,
 		liveStreamRelay,
+		turnLiveStreamRelay,
+		// ponytail: one config-provided In-VM server, nudged by URL; the
+		// orchestration ticket swaps in per-Conversation MicroVM launch here.
+		nudgeInVmServer: inVmServerUrl
+			? async () => {
+					const response = await fetch(`${inVmServerUrl}/nudge`, {
+						method: "POST",
+					});
+					if (!response.ok) {
+						throw new Error(`nudge answered ${response.status}`);
+					}
+				}
+			: undefined,
 		liveStreamTelemetry,
-		closeLiveResources: () => liveStreamRelay.close(),
+		closeLiveResources: async () => {
+			await Promise.all([liveStreamRelay.close(), turnLiveStreamRelay.close()]);
+		},
 		exposureGate,
 		gatewayUpstreamFetch: fetch,
 	};
--- a/apps/chat-api/src/deps.ts
+++ b/apps/chat-api/src/deps.ts
@@ -152,6 +152,7 @@ export function createDeps(
 						new URL("/nudge", config.inVmServerUrl),
 						{
 							method: "POST",
+							signal: AbortSignal.timeout(5_000),
 						},
 					);
 					if (!response.ok) {
--- a/apps/chat-api/src/deps.ts
+++ b/apps/chat-api/src/deps.ts
@@ -76,10 +76,7 @@ export interface AppDeps {
 	 * so the v2 message POST answers 503. Today a dev-mode stub over
 	 * `IN_VM_SERVER_URL`; the orchestration ticket replaces it.
 	 */
-	nudgeInVmServer?: (ref: {
-		userId: string;
-		conversationId: string;
-	}) => Promise<void>;
+	nudgeInVmServer?: () => Promise<void>;
 	/** Cardinality-safe, payload-free Live Stream relay observability. */
 	liveStreamTelemetry: LiveStreamTelemetry;
 	/** Close the lazy Redis relay clients during service shutdown. */
@@ -137,7 +134,6 @@ export function createDeps(
 		url: config.redisUrl,
 		deployment: "current",
 	});
-	const inVmServerUrl = config.inVmServerUrl;
 	return {
 		config,
 		createHarnessChatAgent,
@@ -153,9 +149,9 @@ export function createDeps(
 		turnLiveStreamRelay,
 		// ponytail: one config-provided In-VM server, nudged by URL; the
 		// orchestration ticket swaps in per-Conversation MicroVM launch here.
-		nudgeInVmServer: inVmServerUrl
+		nudgeInVmServer: config.inVmServerUrl
 			? async () => {
-					const response = await fetch(`${inVmServerUrl}/nudge`, {
+					const response = await fetch(`${config.inVmServerUrl}/nudge`, {
 						method: "POST",
 					});
 					if (!response.ok) {
--- a/apps/chat-api/src/deps.ts
+++ b/apps/chat-api/src/deps.ts
@@ -152,6 +152,14 @@ export function createDeps(
 						new URL("/nudge", config.inVmServerUrl),
 						{
 							method: "POST",
+							// A real MicroVM endpoint authenticates with the platform's
+							// per-VM token and routes by port (the image listens on 8080).
+							headers: config.inVmServerAuthToken
+								? {
+										"x-aws-proxy-auth": config.inVmServerAuthToken,
+										"x-aws-proxy-port": "8080",
+									}
+								: {},
 							signal: AbortSignal.timeout(5_000),
 						},
 					);
--- a/apps/chat-api/src/features/conversation-messages/conversation-messages-store.ts
+++ b/apps/chat-api/src/features/conversation-messages/conversation-messages-store.ts
@@ -55,3 +55,34 @@ export interface ConversationMessagesStore {
 		input: ConversationMessagesPageInput,
 	): Promise<ConversationMessagesPage | null>;
 }
+
+export interface TurnRef {
+	userId: string;
+	conversationId: string;
+	messageId: string;
+}
+
+export type EnqueueTurnResult =
+	/** The row was inserted as a `queued` Turn. */
+	| { outcome: "queued" }
+	/** The client message id already names a Turn; nothing changed. */
+	| { outcome: "duplicate"; status: TurnStatus }
+	| { outcome: "not_found" }
+	| { outcome: "archived" };
+
+/**
+ * The /v2 submission contract (ticket #667): chat-api's only write to
+ * `conversation_messages` is the `queued` INSERT; the In-VM server owns every
+ * status transition.
+ */
+export interface TurnSubmissionStore {
+	/**
+	 * Admit a user message as a `queued` Turn under the Conversation row lock,
+	 * so an Archive cannot slip between the check and the insert (the row lock
+	 * is the one `PATCH` takes). Missing and foreign Conversations look
+	 * identical.
+	 */
+	enqueueTurn(input: TurnRef & { parts: unknown }): Promise<EnqueueTurnResult>;
+	/** The Turn's current status, or null when no such Turn exists. */
+	getTurnStatus(ref: TurnRef): Promise<TurnStatus | null>;
+}
--- a/apps/chat-api/src/features/conversation-messages/conversation-messages-store.ts
+++ b/apps/chat-api/src/features/conversation-messages/conversation-messages-store.ts
@@ -54,6 +54,17 @@ export interface ConversationMessagesStore {
 	getPage(
 		input: ConversationMessagesPageInput,
 	): Promise<ConversationMessagesPage | null>;
+	/**
+	 * The /v2 submission write (ticket #667) — chat-api's only write to
+	 * `conversation_messages`; the In-VM server owns every status transition.
+	 * Admits a user message as a `queued` Turn under the Conversation row lock,
+	 * so an Archive cannot slip between the check and the insert (the row lock
+	 * is the one `PATCH` takes). Missing and foreign Conversations look
+	 * identical.
+	 */
+	enqueueTurn(input: TurnRef & { parts: unknown }): Promise<EnqueueTurnResult>;
+	/** The Turn's current status, or null when no such Turn exists. */
+	getTurnStatus(ref: TurnRef): Promise<TurnStatus | null>;
 }
 
 export interface TurnRef {
@@ -69,20 +80,3 @@ export type EnqueueTurnResult =
 	| { outcome: "duplicate"; status: TurnStatus }
 	| { outcome: "not_found" }
 	| { outcome: "archived" };
-
-/**
- * The /v2 submission contract (ticket #667): chat-api's only write to
- * `conversation_messages` is the `queued` INSERT; the In-VM server owns every
- * status transition.
- */
-export interface TurnSubmissionStore {
-	/**
-	 * Admit a user message as a `queued` Turn under the Conversation row lock,
-	 * so an Archive cannot slip between the check and the insert (the row lock
-	 * is the one `PATCH` takes). Missing and foreign Conversations look
-	 * identical.
-	 */
-	enqueueTurn(input: TurnRef & { parts: unknown }): Promise<EnqueueTurnResult>;
-	/** The Turn's current status, or null when no such Turn exists. */
-	getTurnStatus(ref: TurnRef): Promise<TurnStatus | null>;
-}
--- a/apps/chat-api/src/features/conversation-messages/conversation-messages-store.ts
+++ b/apps/chat-api/src/features/conversation-messages/conversation-messages-store.ts
@@ -78,5 +78,7 @@ export type EnqueueTurnResult =
 	| { outcome: "queued" }
 	/** The client message id already names a Turn; nothing changed. */
 	| { outcome: "duplicate"; status: TurnStatus }
+	/** The client message id already names an assistant message (ids are visible in history). */
+	| { outcome: "not_a_turn" }
 	| { outcome: "not_found" }
 	| { outcome: "archived" };
--- a/apps/chat-api/src/features/conversation-messages/conversation-messages.route.test.ts
+++ b/apps/chat-api/src/features/conversation-messages/conversation-messages.route.test.ts
@@ -186,6 +186,9 @@ class FakeTurnStore implements ConversationMessagesStore {
 			return { outcome: "not_found" };
 		}
 		if (this.archived) return { outcome: "archived" };
+		if (input.messageId.startsWith("assistant-")) {
+			return { outcome: "not_a_turn" };
+		}
 		const existing = this.rows.get(input.messageId);
 		if (existing) return { outcome: "duplicate", status: existing.status };
 		this.rows.set(input.messageId, { status: "queued", parts: input.parts });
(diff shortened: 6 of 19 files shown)


chengchao:
## Review of head commit `b112c72` (diff vs `origin/main` = `cbd3521`)

Independent two-axis review: **Standards** (repo's documented coding standards) and **Spec** (issue #667, parent #654). Every finding below was verified against the code at HEAD; what was checked is noted inline.

Verification run locally: `apps/chat-api` conversation-messages tests 25/25 pass; `bunx tsc --noEmit` clean in `apps/chat-api` (apart from the pre-existing `s3-artifact-download-signer.ts` `@smithy/types` error on `main`) and in `packages/agent-db`; `biome check` clean on the touched files.

### Spec

**1. BLOCKING — the Live Stream channel is keyed by the client-chosen message id alone, so Turns in different Conversations (and different owners) with the same message id share one stream.**
- `packages/live-text/src/turn-live-stream-relay.ts:152-153` — channel is `` `${prefix}:{${turnId}}:live` ``, `turnId` only.
- `packages/agent-db/src/schema.ts:224` — `conversation_messages` PK is `(user_id, conversation_id, message_id)`, so the same `message_id` is a *new* `queued` row in any other Conversation, not a duplicate.
- `apps/in-vm-server/src/turn-serving.ts:88` publishes on `claimed.messageId`; `conversation-messages.route.ts:191` subscribes on `ref.messageId`; the route accepts any path-safe id as the Turn id (`route.ts:35`, `:165`).
- Consequence: member B POSTs into their own Conversation with `id` equal to member A's in-flight message id → B's row is admitted `queued` → B's response attaches to A's channel and receives A's chunks; when B's Turn is claimed, B's chunks land on A's stream too, and the first terminal chunk ends both. The same-owner variant is not even adversarial: the doc's own local check uses `"id":"m1"` (`docs/agents/chat-api.md:226`), so any client with per-Conversation ids (m1, m2, …) cross-wires its own Conversations. `docs/agents/security.md` requires every Conversation resource to be owner-scoped, and #667 says the SSE is "scoped to the submitted message's own Turn".
- The keying was chosen on #658, but this route is the first subscriber, so this PR is where the leak becomes reachable. Lazy fix: key the channel on Conversation id + message id (publisher and subscriber both know both; the In-VM server is per-Conversation) and pin it in the relay contract test. Alternatively, have chat-api mint/enforce a globally unique Turn id — but the PK says it isn't one today.

**2. should-fix — the terminal watch can drop the Outcome chunk.**
- `apps/in-vm-server/src/turn-serving.ts:131-132`: `terminalize` (status flip) commits *before* `publish(finish|abort|error)` — the #654 commit-before-publish rule.
- `conversation-messages.route.ts:233-237`: a keepalive tick that reads the row in that window calls `readAbort.abort()` → `AsyncQueue.end()`; a push after `end` is discarded (`packages/live-text/src/live-stream-relay.ts`, `push`: `if (this.#ended) return;`). The client then gets `data: [DONE]` with no terminal chunk (no `finish` / `messageMetadata`), contradicting "ending at its terminal with the Outcome" and AC1's "ends with the Outcome". Window = one DB commit → Redis publish, sampled every 5 s: rare, but structural. Lazy fix: end on the watch only after two consecutive terminal reads (or a one-tick grace), so the relay's own terminal chunk stays the normal ending.

**3. nit / no change — `409` for archived vs #667's "no 409 anywhere".** Verified: v1 `ai-chat.route.ts:105-106` returns `409 { error: "Conversation is archived" }` and #654 scopes "No 409" to the second POST. The author's reading is the right one; suggest ticking the AC with that note.

Verified OK (no finding): stock `DefaultChatTransport` body `{id, messages, trigger, messageId}` (`ai/dist/index.js:18208-18215`; `messageId: undefined` drops under `JSON.stringify`) passes the `.strict()` schema; default `generateId()` ids match `CONVERSATION_ID_PATTERN`; `UI_MESSAGE_STREAM_HEADERS` + `data: [DONE]` match the SDK writer/parser; `POST /nudge` is body-less 202/503 on the In-VM side (`apps/in-vm-server/src/app.ts:43-46`) and `deps.ts` matches; the exposure gate runs after identity and before the write per `security.md`; subscribe-before-nudge (`route.ts:191` → `:205`), client disconnect dropping only the subscription, keepalives while queued, and the duplicate paths are all present and tested with the real relay and the real `ai@7.0.83` parser. The 410-for-terminal-duplicate is a new status a stock transport throws on, but it is documented and within the ticket's latitude.

### Standards

**1. should-fix — keepalive tick can re-arm after the stream has ended** (`conversation-messages.route.ts:229-246`, `:264`). hono's `StreamingApi.write` swallows write errors (`hono/dist/utils/stream.js`), so the `catch` at `:239` never fires on a closed response. If a tick is in flight (awaiting `stream.write` or `getTurnStatus`) at the moment the `for await` loop ends, `finally`'s `clearTimeout(timer)` clears the already-fired id and the tick's trailing `tick()` at `:243` schedules a fresh timer: after a client disconnect this keeps polling `getTurnStatus` every 5 s per dropped client until the Turn is terminal. Bounded, but a leak. One-line fix: `if (readSignal.aborted) return;` before the re-arm at `:243` (or a `stopped` flag set in `finally`).

**2. nit — test does not pin what its name claims** (`postgres-conversation-messages-store.test.ts:248`, "getTurnStatus ignores assistant rows"). An assistant row's `status` is NULL by the DB check, so `selectTurnStatus` returns `null` with or without the `role = 'user'` filter at `postgres-conversation-messages-store.ts:99-104`; the test passes either way and does not protect the `as TurnStatus` cast at `:109` (`CODING_STANDARDS.md`: tautological tests). Either drop it or assert through a path where the filter matters.

**3. nit — nudge `fetch` has no timeout** (`apps/chat-api/src/deps.ts:151-156`) and the route awaits it before sending headers (`route.ts:205`), so a hung In-VM server hangs the POST indefinitely. Dev-mode stub, but `signal: AbortSignal.timeout(…)` is one line.

**4. nit — possible Duplicated Code (judgement call):** `LIVE_STREAM_KEEPALIVE_MS = 5_000` re-declared at `route.ts:20` next to `conversations.route.ts:40`. Defensible as feature-local; mention only.

No documented-standard violations found otherwise: strict body schemas, identity from headers, gate placement, no KB/E2B credentials in `ApiConfig`, Biome style, `@electric-sql/pglite` peer set unchanged, agent-db schema ownership respected (`enqueueTurnTx` only widened to `DbTx`).

### Verdict

Spec finding 1 is a verified cross-Conversation (and cross-owner) Live Stream leak reachable through this route; Spec finding 2 and Standards finding 1 are verified races with one-line fixes.

CHANGES REQUIRED

chengchao:
Addressed the CHANGES REQUIRED review on `b112c72` in follow-up commit `f576a6e` (the current head).

**Spec**
1. Blocking — channel keyed on message id alone: fixed in `packages/live-text`. `openPublisher`/`subscribe` now take `{ conversationId, messageId }`, the channel is `<prefix>:{<conversationId>}:<messageId>:live`, both fields validated path-safe; a new contract test (in-memory + Redis) subscribes the same message id under two Conversations and asserts each sees only its own publisher. The In-VM publisher (`turn-serving.ts`) and the chat-api subscriber pass the pair.
2. Should-fix — terminal watch dropping the Outcome chunk: the watch now ends the stream only after two consecutive terminal reads (10 s), so the single commit-before-publish window can no longer win. Documented in `docs/agents/chat-api.md`.
3. Nit — `409` for archived: kept, as agreed.

**Standards**
1. Should-fix — tick re-arm after the stream ends: the tick re-arms only while the read signal is live (`if (!readSignal.aborted) tick()`).
2. Nit — tautological role-filter test: the redundant `role = 'user'` filter is gone (the schema check already pins an assistant row's status to NULL) and so is the test.
3. Nit — nudge `fetch` without a timeout: `AbortSignal.timeout(5_000)` added.
4. Nit — duplicated keepalive constant: left as is.

Verification after the changes: live-text 53 pass, in-vm-server 93 pass, chat-api conversation-messages + config 42 pass; `tsc` and biome clean on all three (chat-api's pre-existing signer error aside).

chengchao:
## Review of head commit `f576a6e` (diff vs `origin/main` = `cbd3521`)

Fresh two-axis review (Standards / Spec #667, parent #654), independent of the earlier one. Every finding below was verified against the code at `f576a6e`; nothing is taken from the PR body or the author's reply on trust.

Verification run locally: `packages/live-text` 53/53, `apps/in-vm-server` 93/93, `apps/chat-api` conversation-messages + config 42/42 pass; `bunx tsc --noEmit` clean in live-text, in-vm-server, agent-db, and chat-api apart from the pre-existing `s3-artifact-download-signer.ts` error on `main`; `biome check` clean on all 15 touched `.ts` files. CI runs each workspace's own `test` script (`scripts/test-workspaces.ts`), so the 10 s terminal-watch test gets chat-api's `--timeout=30000`.

### Previous review (on `b112c72`) — status of each finding

| # | Finding | Status at `f576a6e` |
|---|---|---|
| Spec 1 (blocking) | Live Stream channel keyed on message id alone | **Resolved.** `turn-live-stream-relay.ts:155-157` keys on `{conversationId}:{messageId}`; both fields validated path-safe (`live-stream-validation.ts:37-48`); publisher (`turn-serving.ts:88-91`) and subscriber (`route.ts:191`) pass the pair; the cross-talk contract test (`turn-live-stream-relay.contract.ts:32-55`) runs on both the in-memory and Redis transports. Conversation ids are server-minted UUIDs (`conversations.controller.ts:46`), so the pair is effectively global. |
| Spec 2 (should-fix) | Terminal watch could drop the Outcome chunk | **Resolved.** `route.ts:236-242` ends the stream only on the second consecutive terminal read; the In-VM commit-then-publish window (`turn-serving.ts:131-135`) is a single read. Turn status is monotonic, so the counter needs no reset. Documented in `docs/agents/chat-api.md`. |
| Spec 3 (nit) | `409` for archived vs "no 409 anywhere" | Kept, as agreed; see Spec nit 3 below. |
| Standards 1 (should-fix) | Keepalive tick re-arms after the stream ended | **Resolved.** `route.ts:247` re-arms only `if (!readSignal.aborted)`; `finally` at `:267-270` aborts `readAbort` so an in-flight tick cannot re-arm. |
| Standards 2 (nit) | Tautological role-filter test | **Resolved.** The `role = 'user'` filter and its test are gone; `postgres-conversation-messages-store.ts:106-109` explains why the status column alone suffices. |
| Standards 3 (nit) | Nudge `fetch` without a timeout | **Resolved.** `deps.ts:155` — `signal: AbortSignal.timeout(5_000)`. |
| Standards 4 (nit) | Duplicated keepalive constant | Left as is, as stated. |

The author's reply matches the code on every point.

### Standards

**1. should-fix — a status-read failure in the tick silently truncates a healthy stream, with no log line.** `conversation-messages.route.ts:243-246`: `catch { stream.abort(); return; }`. hono 4.12's `StreamingApi.write` swallows errors (`hono/dist/utils/stream.js`), so the only thing that can throw inside the tick is `getTurnStatus`. On a transient DB error the tick calls `stream.abort()`, which sets `stream.aborted` → `onAbort` → `readAbort.abort()` → the relay queue ends cleanly → the `for await` at `:252` exits normally → `gone()` is true → neither the `[DONE]` at `:257` nor the `logger.error` at `:262` runs. The client's live stream ends mid-Turn while the relay is still delivering, and nothing is logged. The v1 keepalive precedent aborts only on a *write* failure (client gone) and logs it (`conversations.route.ts:97-111`). Suggested: on a failed status read, log and skip the read (re-arm), rather than abort — the watch is a safety net, not the stream's liveness signal.

**2. nit — 22 s unit file.** `LIVE_STREAM_KEEPALIVE_MS = 5_000` (`route.ts:20`) is not injectable, so the terminal-watch test (`route.test.ts:493`) waits two real ticks and "two rapid POSTs" (`:400`) 5.4 s; measured ~22 s for the 15 route tests. Same trade-off as the v1 keepalive test; mention only.

**3. nit — tick may outlive the response.** `route.ts:267-270` — `clearTimeout` on an already-fired timer is a no-op, so an in-flight tick issues one more `getTurnStatus` after the stream closed. Harmless (aborts are idempotent), one wasted query.

**4. nit — smells (judgement calls).** *Duplicated Code*: `AbortSignal.any([c.req.raw.signal, readAbort.signal])` (`route.ts:188`) re-does v1's `linkedAbortController` (`conversations.route.ts:89-94`); defensible since v1 retires. *Feature Envy*: `isTerminalTurnStatus` (`route.ts:51-53`) casts agent-db's tuple — belongs beside `TurnOutcome` in `turn-store.ts`. *Vocabulary*: `route.ts:122` "never cancels anything" — CONTEXT.md avoids "cancel" for Interruption.

No documented-standard violations found: identity from `X-*` headers, `.strict()` bodies, exposure gate after identity and before the INSERT (`route.ts:158-170`) with the exact `403` text, no KB/E2B credentials in `ApiConfig` (`env.test.ts` still pins it), `@electric-sql/pglite` peer already present, agent-db schema ownership respected, `ai` pinned to `7.0.83` matching live-text and in-vm-server, Biome style. No tautological tests found beyond Spec nit 2 below.

### Spec

**1. should-fix — client-triggerable 500 on a duplicate id that names an assistant row.** Spec #654: "Re-POST of the same client message id is a no-op (PK)". The PK is `(user_id, conversation_id, message_id)` with no role (`packages/agent-db/src/schema.ts:224`), and assistant ids are visible to the owner in `GET /v2/conversations/:id/messages` (`postgres-conversation-messages-store.ts:116`). A POST whose message `id` equals one of those: `enqueueTurnTx`'s `onConflictDoNothing` returns false (`turn-store.ts:68-73`), `selectTurnStatus` returns `null` because the check constraint pins an assistant row's status to NULL (`schema.ts:241`), and `postgres-conversation-messages-store.ts:80-83` throws → unhandled → hono `500`. Nothing is written and only the owner can trigger it, but it is a client-input condition answered as a server error and it is not covered by the store test (`postgres-conversation-messages-store.test.ts:190-247`). One-line fix: return a 4xx-shaped outcome (e.g. treat it as `duplicate` with a `400`/`409` mapping, or a new `not_a_turn` outcome) and pin it in the store test.

**2. nit — the "two consecutive terminal reads" rule is not pinned by a test.** `route.test.ts:493-516` sets `interrupted` before the first tick and asserts only `pings >= 1` and `["[DONE]"]`; a regression to `terminalReads >= 1` at `route.ts:239` passes identically, just 5 s sooner. Assert `pings >= 2`.

**3. nit — AC6 test cannot detect a regression.** `route.test.ts:473-491`: the fake drain (`:230-255`) finishes regardless of what the route does, and the route has no cancel path to regress, so the test proves non-crash only (`CODING_STANDARDS.md`). The AC is carried by the live run in the PR body. Acceptable; mention only.

**4. nit — `409` for archived vs ticket's "no 409 anywhere".** `route.ts:174-176` matches v1 (`conversations.route.ts:515`) and #654 scopes "No 409" to the second POST; the reading is right, but tick the AC with that note or amend the ticket text.

**5. nit — exposure gate on the re-attach path diverges from `security.md:20`.** "Reconnect … for existing owned resources do not consult the exposure gate", and v1 skips the gate on an exact retry (`conversations.route.ts:490-495`); `route.ts:158` gates before `enqueueTurn`, so a duplicate attach is gated. Fail-closed and declared in the PR body — either amend `security.md` or match v1.

**6. note, not a PR deviation — persist→subscribe window.** The drain loop chains queued rows without a nudge (`drain-loop.ts:103`), and the INSERT commits (`route.ts:167`) before `subscribe` (`:191`), so a predecessor finishing during the Redis subscribe round-trip can cost the successor its first chunks. #654 orders it exactly this way (persist → subscribe → nudge, no backlog); recording it so the orchestration ticket knows.

**7. note — unbounded stream when the Turn is never served.** The dev stub nudges one fixed VM regardless of Conversation (`deps.ts:150-163`); a POST for any other Conversation stays `queued` and the tick never terminates. Inherent to the ticket's "dev-mode stub"; no spec bound exists.

Scope: the relay re-key across live-text / in-vm-server, `ai` as a chat-api dependency, `DbTx` in `enqueueTurnTx`, and the `MAX_MESSAGE_LENGTH` export are all necessary for the route; no unrequested behaviour found. `docs/agents/chat-api.md` matches the code (5 s pings, `410` + `recovery: "history"`, `403` before any write, `409` archived, two-read terminal watch).

Per-AC: 1 met (live claim; publisher/subscriber key and chunk grammar consistent) · 2 met (`route.test.ts:353-382`, real `DefaultChatTransport` + `readUIMessageStream`) · 3 met (`:400-429`, exact `toEqual` proves zero leakage) · 4 met (`:384-398`, synchronous publish inside the nudge) · 5 met with the Spec 1 hole · 6 met by live claim, test is non-detecting (Spec 3) · 7 met (`:518-551`).

### Verdict

All six of the previous review's findings are resolved as the author described. Two new verified should-fix findings remain, each a one-line fix: a transient status-read failure silently truncates a healthy stream without a log (Standards 1), and a re-POST reusing an assistant message id is a client-triggerable 500 (Spec 1).

CHANGES REQUIRED

chengchao:
Addressed the second CHANGES REQUIRED review (on `f576a6e`) in follow-up commit `9c3d299` (the current head).

**Standards**
1. Should-fix — watch read failure ended the stream: the tick now separates the two operations. A failed `: ping` write still aborts (the client is gone); a failed `getTurnStatus` logs `v2 terminal watch read failed` and retries next tick, the response staying open. The terminal counter also resets on a non-terminal read, so "two consecutive" is literal.
2. Nit — non-injectable interval: left as is (v1 precedent, same 5 s constant).
3. Nit — extra read after close: the tick now returns immediately when the read signal is already aborted.
4. Nit — smells: "cancels" → "interrupts" in the route doc comment; the rest left as is.

**Spec**
1. Should-fix — assistant-id reuse → 500: `enqueueTurn` now reports `not_a_turn` when the PK conflict is with an assistant row, and the route answers `409 { error: "Message id names an assistant message" }` with nothing written. Covered in the pglite store test and the route test; documented in `docs/agents/chat-api.md`.
2. Nit — the terminal-watch test asserts `pings >= 2` now.
3. Nit — AC6 disconnect test: left as is (the live run carries the AC, as noted).
4. Nit — `409` archived: unchanged.
5. Nit — gate on a duplicate re-attach: left as is, deliberately — the gate runs before any read or write, and denying a re-POST to a gated-off member fails closed; noted in the PR body.

Verification: chat-api conversation-messages 25 pass (15 route, 10 store); `tsc` and biome clean (pre-existing signer error aside).

chengchao:
## Review of head commit `9c3d299` (diff vs `origin/main` = `cbd3521`)

Fresh two-axis review (Standards / Spec #667, parent #654), independent of the two earlier rounds. Every claim below was checked against the code at `9c3d299`; nothing is taken from the PR body or the author's replies on trust.

Verification run locally: `packages/live-text` 53/53, `apps/in-vm-server` 93/93, `apps/chat-api` conversation-messages 25/25 (15 route, 10 store) pass; `bunx tsc --noEmit` clean in live-text, in-vm-server, agent-db, and chat-api apart from the pre-existing `s3-artifact-download-signer.ts` error on `main`; `biome check` clean on all 15 touched `.ts` files.

### Previous review (on `f576a6e`) — status of each finding

| # | Finding | Status at `9c3d299` |
|---|---|---|
| Standards 1 (should-fix) | A failed status read silently truncated a healthy stream | **Resolved.** `route.ts:236-263`: the ping write and the status read are separate `try` blocks; a read failure hits `logger.warn("v2 terminal watch read failed")` and the tick re-arms (`:264`), the response stays open. The counter resets on a non-terminal read (`:245-248`), so "two consecutive" is literal. |
| Standards 2 (nit) | Non-injectable 5 s interval | Left as is, as stated. |
| Standards 3 (nit) | Extra `getTurnStatus` after close | **Resolved.** `route.ts:235` returns before any work when `readSignal.aborted`. |
| Standards 4 (nit) | "cancels" vocabulary | **Resolved.** `route.ts:122` reads "chat-api never interrupts anything". |
| Spec 1 (should-fix) | Assistant-id reuse → unhandled `500` | **Resolved.** Store returns `not_a_turn` when the PK conflict finds a NULL status (`postgres-conversation-messages-store.ts:80-83`); route answers `409 { error: "Message id names an assistant message" }` before any subscribe/nudge (`route.ts:177-179`); pinned in the pglite store test (`:243-250`) and the route test (`:475-484`, asserts no nudge); documented in `docs/agents/chat-api.md`. |
| Spec 2 (nit) | Two-read rule not pinned | **Resolved.** `route.test.ts:528` asserts `pings >= 2`. |
| Spec 3/4/5 (nits) | AC6 test shape, `409` archived, gate on re-attach | Left as is, as stated. |

The author's reply matches the code on every point.

### Standards

No documented-standard violations found. Checked: identity from `X-*` headers only; top-level body and the submitted message are `.strict()`; the exposure gate runs after identity and before the only write (`route.ts:158-170`, test pins `rows.size === 0` on `403`); `ApiConfig` gains only a non-secret URL and `env.test.ts` still pins no KB/E2B secrets; `ai@7.0.83` matches the live-text / in-vm-server pin and no new `drizzle-orm` consumer appears (pglite peer set untouched); `enqueueTurnTx` is only widened to `DbTx`; Biome clean. Concurrency in the SSE tick verified against hono 4.12.12 source: the `finally` at `route.ts:284-287` aborts `readAbort`, every re-arm is guarded by `!readSignal.aborted`, the relay ends the iteration via `queue.end()` on abort (no throw) so the terminal-watch exit still reaches `[DONE]` (`gone()` checks `stream.aborted`, not `readSignal`), and the subscription's listener is removed on close. No leaks found.

Nits (none blocking):

1. **nit — unreachable branch.** `route.ts:238-241` `catch { stream.abort(); return; }`: hono's `StreamingApi.write` wraps `writer.write` in `try {} catch {}` and never rejects (`node_modules/hono/dist/utils/stream.js`), so a write can't fail here. Client disconnect is actually detected by `responseReadable.cancel → abort() → onAbort → readAbort.abort()` (and `c.req.raw.signal`), which the tick honours at `:235` — behaviour is correct, but the "failed write aborts" story in the reply/comment is dead code. v1's `startAgUiKeepalive` has the same dead `.catch`; mirrored precedent.
2. **nit — body strictness (judgement).** `route.ts:47` `messages: z.array(z.unknown()).min(1)`: every non-final message is unvalidated JSON up to the 10 MB body cap, where the v1 harness route used a one-element `z.tuple`. Deliberate for the stock `useChat` body and documented in `chat-api.md`; mention only.
3. **nit — possible Duplicated Code.** `selectTurnStatus` (`postgres-conversation-messages-store.ts:92-107`) rebuilds agent-db's unexported three-`eq` `turnKey()` (`turn-store.ts:33-43`), whose header already says "chat-api only enqueues and reads" — a `getTurnStatusTx` beside `enqueueTurnTx` would be the natural home. Also `AbortSignal.any([...])` at `route.ts:191` re-does v1's `linkedAbortController`; defensible since v1 retires.
4. **nit — possible Data Clump.** `TurnRef` (chat-api) and `LiveStreamTurnKey` (live-text) are two types for the same `{conversationId, messageId}` pair; `subscribe(ref, …)` only works structurally.
5. **nit — wall clock.** The 15 route tests take ~22 s locally (two tests wait for real 5 s ticks). Same trade-off as v1; CI runs chat-api's own `--timeout=30000`.

### Spec

No missing or wrong requirement found. Traced: duplicate POST on a `queued`/`processing` Turn attaches and nudges (`route.ts:180-186` falls through; the nudge is idempotent via `claimNextTurnTx`); terminal duplicate → `410 { recovery: "history" }` with no nudge; subscribe failure → best-effort nudge then `503` (`:195-203`); nudge failure → warn and stream on (`:204-216`, the drain loop's 15 s self-heal covers it); the In-VM server terminalizes before publishing the terminal chunk (`turn-serving.ts:131-135`) and the route ends only on the second consecutive terminal read, then writes `[DONE]`; a publisher that closes without a terminal chunk fails the subscriber and the route closes without `[DONE]` (Recovery from history); client disconnect ends only chat-api's subscription. `docs/agents/chat-api.md` "Submit a v2 message" matches the code line for line (order of operations, 5 s pings, two-read watch, `410`/`409` payloads).

Per-AC: 1 met by the PR body's live transcript (no artefact in the branch; the doc's local-check block is the runbook) · 2 met (`route.test.ts:356-385`, real `DefaultChatTransport` + `readUIMessageStream` on `ai@7.0.83`) · 3 met (`:403-432`, exact `toEqual` per stream, `pings >= 1`) · 4 met (`:387-401`, synchronous publish inside the nudge — would lose chunks if reordered) · 5 met (`:434-485` + store `:199-250` + docs) · 6 met (`:487-505`, fake store row rather than `GET` history; the live run carries the durable-history half) · 7 met (`:532-565`; `409` archived is the v1 code and #654 scopes "no 409" to the second POST).

Nits (none blocking):

1. **nit (doc) — the attach path is not stock-client-consumable mid-part.** A re-POST that attaches while the In-VM server is inside a text (or tool-input) part delivers a `text-delta` with no prior `text-start`; the stock parser throws `UIMessageStreamError` on it (`node_modules/ai/dist/index.js:7069-7076`; same for `text-end`, `tool-input-delta`). The behaviour is the spec's own "no backlog" choice and the doc already says earlier chunks are missed; only the general sentence "A stock `useChat` needs no custom transport" (`chat-api.md:198`) is over-broad for that path. One added clause — that a mid-part attach ends in the client's `onError` and Recovery from history — would close it. Nothing in `useChat`/`DefaultChatTransport` re-POSTs on its own, so only a deliberate client retry reaches this.
2. **note — unbounded stream for a never-served Turn.** The dev stub nudges one fixed VM; a Turn that is never claimed keeps the response open with pings indefinitely (`route.ts:233-267` has no ceiling). Inherent to the ticketed stub; no spec bound exists. Carry to the orchestration ticket.
3. **note — cross-ticket scope, accepted.** The relay re-key (`packages/live-text`, `in-vm-server/turn-serving.ts:88-91`) amends #658's contract and the `409` for an assistant id is a status #667 didn't define; both were required by earlier review findings and are documented.

### Verdict

All six of the previous review's findings are resolved exactly as the author described. The new findings are nits and notes only — one unreachable branch, a doc sentence that is over-broad for the mid-part attach path, and smell-level judgement calls — none is a defect against #667 or a documented standard.

APPROVED FOR MERGE

chengchao:
## Real-topology verification (2026-09-01, head `658e495`)

Ran the #667 route against the production topology, not just the local stack: a scratch MicroVM image (`mymemo-agent-verify-667`, built from this branch by `scripts/deploy/register_microvm_image.sh` — needed because the PR re-keys the Turn Live Stream channel, so the prod image's In-VM server publishes where this chat-api no longer listens), one hand-launched VM through the prod VPC egress connector (`microvm-a576899f…`, RUNNING, `/health` ok through the JWE endpoint), and a one-off Fargate task (`22bdee40…`, chat-api's subnets/SGs/roles) running chat-api's local composition built from this branch (`IN_VM_SERVER_URL` = the VM endpoint, `IN_VM_SERVER_AUTH_TOKEN` = the per-VM token, prod `AGENT_DATABASE_URL`/`DB_PASSWORD`/`REDIS_URL`). The task's e2e script seeded one synthetic Conversation for a synthetic member, drove `POST /v2/conversations/:id/messages` on `127.0.0.1:3000`, and read history back.

**E2E PASS — 14/14** (task exit 0):

| check | evidence |
| --- | --- |
| redis subscriber survives 75 s idle (Valkey, TLS) | message received after the idle window, no client events |
| AC1 status 200 + UI stream headers | `text/event-stream`, `x-vercel-ai-ui-message-stream: v1` |
| AC1 chunk sequence start…finish + `[DONE]` | `start, start-step, text-start, text-delta×3, text-end, finish-step, finish{status:done}`, then `[DONE]`; 14.5 s end to end on a cold VM (2 keepalives while the CLI spun up and called the gateway) |
| AC1 durable history matches the stream | user row `done`, assistant row present with the same text |
| AC5 duplicate of a terminal Turn | `410 {"error":"Turn already ended","recovery":"history"}` |
| AC3 two rapid POSTs | both 200; the second held with 2 keepalives, then only its own Turn's chunks; the first complete and unpolluted |
| AC6 client disconnect mid-stream | aborted after the first frame; history later shows the Turn `done` with the full 21-line answer |
| AC7 foreign owner | 404 |
| AC7 archived Conversation | PATCH 200 → POST `409 {"error":"Conversation is archived"}` |
| cleanup | DELETE 204, zero rows left |

The model calls went VM → internal ALB → `/v2/gateway/<conversation>` on the minted token (attribution lines in chat-api's log group). Torn down afterwards: VM terminated, taskdef deregistered, ECR tag and scratch MicroVM image deleted, secret files shredded.

Two earlier attempts, for the record: run 1 died on a transient `GetMicrovm` 502 (the launch script now retries every post-launch read); run 2 ran against the **prod** MicroVM image and streamed zero chunks — the channel-key mismatch above, which is exactly what the review's re-key finding predicted — and was what motivated the scratch image. Run 2 also surfaced that the prod OpenRouter account had ~$0.40 of credit left (Claude Code's 64k `max_tokens` request was answered 402), fixed by a top-up before run 3.

Follow-up material seen along the way, not this PR's: an OpenRouter 402 reaches the In-VM mapper as an assistant message outside a provider call, so the Turn ends `error` with the provider's reason only in the VM's log.

chengchao:
## Review of head commit `32999e4` (diff vs `origin/main` = `cbd3521`)

Fresh two-axis review (Standards / Spec #667, parent #654), independent of the three earlier rounds. Every claim below was checked against the code at `32999e4`; nothing is taken from the PR body, the earlier replies, or the real-topology comment on trust. Focus: the two commits since the `9c3d299` approval — `658e495` (`IN_VM_SERVER_AUTH_TOKEN` on the dev-mode nudge stub) and `32999e4` (keepalive/terminal-watch loop over an abortable `delay`; pino `err`).

Verification run locally: `packages/live-text` 53/53, `apps/in-vm-server` 93/93, `apps/chat-api` conversation-messages + config 43/43 pass; `bunx tsc --noEmit` clean in live-text and in-vm-server, and in chat-api apart from the pre-existing `s3-artifact-download-signer.ts` error on `main`. Bun 1.3.0's `node:timers/promises` `setTimeout` was probed directly: it rejects within ~1 ms of the signal aborting and immediately on an already-aborted signal, and holds no timer after abort.

### Since `9c3d299` — what was checked

**Keepalive / terminal-watch loop** (`apps/chat-api/src/features/conversation-messages/conversation-messages.route.ts:230-262`, `:263-281`), traced against hono 4.12.12 (`StreamingApi.write` swallows errors; `responseReadable.cancel → abort() → onAbort`) and `AsyncQueue` (`packages/live-text/src/live-stream-relay.ts:554-600`, `end()` drains buffered values before the iteration returns):

- Disconnected client: cancel → `stream.abort()` → `readAbort.abort()` (`:220`) → `delay` rejects → `catch` sees `readSignal.aborted` → loop returns; the relay's abort listener ends the queue, the `for await` exits cleanly, `gone()` is true so no `[DONE]`; `finally` aborts again (idempotent). No re-arm, no leaked timer, no extra status read beyond one possibly in flight. Correct.
- Turn ends via the terminal chunk: `queue.end()` on the terminal type → `[DONE]` written → `finally` aborts → the loop's next `delay` rejects → returns. Correct.
- Turn ends only via the watch: second consecutive terminal read → `readAbort.abort()` → queue ends (buffered chunks, a late terminal one included, are still yielded first) → `gone()` false → `[DONE]`. Correct; pinned by `route.test.ts:507-530` (`pings >= 2`, `["[DONE]"]`).
- Failed status read: `logger.warn` and the loop continues, response open. Correct.

**Auth-header branch** (`apps/chat-api/src/deps.ts:151-166`, `config/env.ts:77-84`): headers only when the token is set; names and port match the operator runbook in `infra/microvm-image/README.md:58-66`; the token stays in chat-api's env and never reaches the VM, image, or payload; `env.test.ts:82-90` pins both values and the KB/E2B exclusion still holds.

**pino `err`**: `pinoLogger({ pino: { level } })` builds a stock pino root whose default `serializers.err` is inherited by the request child that `c.var.logger.warn` delegates to; `{ err: error }` matches `ai-chat.route.ts`'s convention.

### Standards

No documented-standard violations found. Identity from `X-*` headers only; `.strict()` bodies; the exposure gate runs after identity and before the only write (`route.ts:159-171`); `ApiConfig` holds no KB or E2B credential; `ai@7.0.83` matches the live-text / in-vm-server pin; no new `drizzle-orm` consumer; Biome clean.

Nits (none blocking):

1. **nit — stale header comment in a touched file.** `apps/chat-api/src/features/conversation-messages/conversation-messages-store.ts:6` still says "chat-api only reads this projection — the In-VM server is the sole writer", which the same file's `enqueueTurn` contract (`:57-64`) and `deps.ts:63` now contradict. One-line fix: qualify it to "chat-api's only write is the `queued` INSERT".
2. **nit — `security.md` drift.** The `ApiConfig` paragraph (`docs/agents/security.md:455`) enumerates the production secrets chat-api may hold (OpenRouter key, gateway secret); `inVmServerAuthToken` is a third bearer-shaped value, documented in `configuration.md` but not there. The boundary itself is respected (chat-api-only, never delivered to a VM); one clause would close the drift.
3. **nit — "two consecutive" is per successful read, not per tick.** `route.ts:242-260`: a failed read neither increments nor resets `terminalReads`, so terminal → read-failed → terminal ends the stream, while `docs/agents/chat-api.md:199` says "two consecutive ticks". Harmless (≥10 s have elapsed by then, well past the commit-before-publish window); either reset in the `catch` or soften the doc.
4. **nit — unreachable branch (carried).** `route.ts:238-241`: the only thrower in that `try` is `delay` on abort, where `readSignal.aborted` is always true, so `stream.abort()` is dead; `catch { return; }` is the honest shape. Same nit as the `9c3d299` round.
5. **nit — untested branch.** The read-failure path (`route.ts:253-260`, the `9c3d299` fix) has no test: the fake store's `getTurnStatus` (`route.test.ts:198`) never throws. A one-off override that throws once and asserts the stream still ends with `[DONE]` would pin it.
6. **nit — smells (judgement calls).** *Duplicated Code*: `LIVE_STREAM_KEEPALIVE_MS` and the abortable-delay shape also live in `conversations.route.ts:40,81-87`; defensible since v1 retires. *Vocabulary*: "Ensure-VM" (`deps.ts:72`, `env.ts:73`, both docs) is not in `CONTEXT.md` — flag for `/domain-modeling` per `domain.md`.

### Spec

No missing or wrong requirement found. Per-AC at `32999e4`: 1 met by the PR body's local run and the real-topology comment (no artefact in the tree; expected for this AC) · 2 met (`route.test.ts:356-385`, real `DefaultChatTransport` + `readUIMessageStream`) · 3 met (`:403-432`) · 4 met (`:387-401`, synchronous publish inside the nudge) · 5 met (`:434-485`, store test, docs) · 6 met (`:487-505`; the test proves non-hang only, the live run carries the property — as noted on earlier rounds) · 7 met (`:532-565`; `409` archived is v1's code and accepted).

Nits / notes (none blocking):

1. **nit — carried, previously declined:** the gate runs before ownership and on duplicate re-attaches (`route.ts:159-177`), where v1 checks 404/409 first (`ai-chat.route.ts:98-110`) and skips the gate for an existing Run (`conversations.route.ts:490-495`; `exposure-gate.ts:22`). `security.md`'s rule — after identity, before any write — is satisfied; fail-closed and declared. Mention only.
2. **note — scope of `658e495`.** #667 words the stub as a "config-provided In-VM server endpoint"; `IN_VM_SERVER_AUTH_TOKEN` lets that stub reach a platform-authenticated MicroVM endpoint, which is orchestration-ticket surface. It is optional, documented, eight lines, and is what made the real-topology verification possible; acceptable, but the orchestration ticket should delete it rather than build on it.
3. **note — documented partial:** a watch-ended stream closes with `[DONE]` and no Outcome chunk (`route.ts:249-251, 269`), so a stock `useChat` sees no `finish`/`abort` for that path and reaches the Outcome via history. Deliberate (no synthesized chunks, per #654) and documented in `chat-api.md:199-203`.

### Verdict

Both post-approval commits do what they say: the loop restructure preserves every abort/close property the `9c3d299` review verified (and drops the timer bookkeeping that needed guarding), and the auth-header branch is a correctly scoped, documented dev-mode affordance. Every finding above is a nit, a doc drift, or a previously declined judgement call — none is a defect against #667 or a documented standard.

APPROVED FOR MERGE

Issue #669, opened 2026-08-31
v2: orchestration — claim, launch, rehydrate


chengchao:
## Parent

Spec #654 — /v2 chat on Lambda MicroVMs.

## What to build

chat-api drives the VM lifecycle inline — no orchestrator service, no background deployable. New `conversation_vm` table (1:1 with Conversation: VM id, endpoint, image version, state, last activity, Checkpoint pointer) doubling as the transactional launch claim: `INSERT … ON CONFLICT DO NOTHING RETURNING` claims the launch; rehydrate re-claims via atomic UPDATE from `terminated`; a stale `launching` claim (~2 min) is re-claimable. Per POST: persist → claim/ensure VM (`RunMicrovm` with SDK adaptive retry and keepalives, retrying the platform's transient 502; retryable 503 to the client on exhaustion) → mint the Gateway token and deliver it via `runHookPayload` → subscribe → nudge (a nudge auto-resumes a suspended VM). Lazy rehydrate only — never proactive; Archive triggers nothing and unarchive does nothing until the next message. Idle policy as config: maxIdle 900 s / suspendedDuration 3600 s / autoResume on / maximum duration 28800 s (the platform enforces the 8 h cap; at-cap kills are handled reactively). Resume never upgrades the image; upgrades land only at rehydrate, plus an urgent flag that converts the next resume into a rehydrate. Replaces the message-POST ticket's dev stub.

## Acceptance criteria

- [x] POST to a Conversation with no VM cold-launches one and the Turn streams end-to-end in staging, through the Gateway.
- [x] Concurrent POSTs to a VM-less Conversation launch exactly one VM — claim race proven in the real-Postgres lane.
- [x] A suspended VM auto-resumes on nudge; nothing resumes proactively.
- [x] After termination, the next POST rehydrates lazily onto the current image (visible via the recorded image version).
- [x] Launch failure after retries surfaces a retryable 503; a stale `launching` claim is re-claimable after ~2 min.
- [x] The urgent-upgrade flag converts the next resume into a rehydrate.

## Blocked by

- #659 — Gateway route
- #660 — Terraform inventory
- #667 — message POST + own-Turn SSE
- #666 — In-VM server inside a real VM

chengchao:
[...]

PR #700, opened 2026-09-02
feat(v2): Checkpoint + lifecycle hooks — suspend drains and checkpoints through chat-api, run restores before ready (#670)

### LINKED ISSUE #670: v2: Checkpoint + lifecycle hooks

chengchao:
Closes #670 (spec #654).

## What

Durability across VM replacement. Built on the issue's two findings: **the VM cannot reach S3** (no VPC endpoint — that absence is the egress lockdown) and **both lifecycle hooks are capped at 60 s**.

**chat-api — the Checkpoint door** (`features/checkpoint/`): `PUT` / `GET /v2/checkpoint/:conversationId`, authenticated by the per-Conversation gateway token — the same token the model gateway verifies. chat-api derives the Conversation from the verified token and writes exactly `conversations/<id>/<uuid>.tar.gz` in the checkpoint bucket: per-Conversation scoping in code, no VM-side IAM at all.
- `PUT` streams the body (Content-Length required, 32 MiB cap — what the 45 s restore budget moves at the platform's quoted low-end bandwidth, raised once measured live; a Node `Readable` piped straight into `PutObject`, never buffered) to a fresh key, moves `conversation_vm.checkpoint_pointer` to it in one write **guarded on the VM id** (`x-mymemo-microvm-id` must be the row's VM — a VM retired by an urgent upgrade while its suspend hook was still draining gets `409` and its object is removed, so it can never fork the lineage), then deletes the previous object best-effort. The pointer always names a complete object: the latest durable Checkpoint.
- `GET` streams the object the pointer names; `204` = nothing to restore; `404` = dangling pointer (the VM fails its boot rather than serve from nothing).
- The gateway token gains a `userId` claim (`{ conversationId, userId, exp }`) so the `conversation_vm` row is addressed by its primary key. Same "ship together" story as #694: a token minted by the previous chat-api is rejected by the new one.
- `MICROVM_CHECKPOINT_BUCKET` joins the `MICROVM_IMAGE_ARN` block; the payload gains `CHECKPOINT_URL`.

**In-VM server — the hooks**:
- `/suspend` is the **graceful-drain gate**: `drainLoop.pause()` refuses new claims and settles once the loop is parked with nothing in flight; then the full Checkpoint (`~/.claude` minus `debug`/`shell-snapshots`, plus the Workspace, one tar with two roots) is packed with the image's `tar` and `PUT`. 200 only once it landed; a failed `PUT` is a 500 and lifts the pause so a VM the platform keeps running keeps serving.
- `/resume` lifts the pause; a pause outlived by resume settles once the queue drains and the loop parks idle, so the Checkpoint is always taken with nothing in flight. A suspend that finds a Turn still `queued` once the Checkpoint lands (one that arrived during the hold, or was waiting behind the Turn in flight — a `hasQueuedTurnsTx` read, not a flag) answers 500 and lifts the pause rather than strand that Turn until the next message; a hold outlived by resume settles only at the next hold, so a Checkpoint is never taken mid-Turn. A payload naming `CHECKPOINT_URL` while the run hook names no `microvmId` fails the launch instead of serving without durability.
- `/run` **restores before anything else**: `GET` → unpack into HOME and the Workspace → boot sweep → drain loop. A restore that fails fails the launch; the client's re-POST rehydrates again.
- **Model-side continuity**: the Agent session id is pinned to the Conversation id (a UUID, as the SDK requires — chat-api mints them with `crypto.randomUUID()`, and the In-VM config now asserts it). A fresh session is created under it (`sessionId`), and whenever its transcript exists under `~/.claude/projects/` — after a restore on a new VM, or a retired session within one process — it is resumed (`resume`). The transcript's directory is keyed on the cwd, so this relies on the image's fixed Workspace path — the probe below confirms it.
- A hard kill mid-Turn (no hook) is unchanged: the replacement VM's boot sweep terminalizes the Turn `interrupted`, before `/run` answers.

**Infra**: `tar` + `gzip` added to the image (AL2023 minimal ships neither — verified in the container), `suspendTimeoutInSeconds=60` at image registration (the same lesson as `runTimeoutInSeconds` on #689), chat-api's task role gains `s3:GetObject`/`PutObject` on `conversations/*`, and the VM execution role's checkpoint policy is **removed** (the `conversations/*` residual ADR-0034 accepted no longer exists). The role itself stays so the launch contract is unchanged; dropping it outright is a follow-up once a live `RunMicrovm` without one is proven. ADR-0034 gains a second amendment.

## Acceptance criteria

- [x] **Suspend→resume preserves the Agent session and Workspace natively** — nothing here touches the snapshot; `/resume` only lifts the drain gate (`app.test.ts`, `drain-loop.test.ts`). Live: needs the real topology (below).
- [x] **A suspend arriving mid-Turn holds until the Turn's terminal, then the Checkpoint completes** — `drain-loop.test.ts` (pause does not settle while a Turn streams, settles at its terminal, a queued successor waits for resume) and `index.ts`'s suspend handler (pause → checkpoint → 200). Live: real topology.
- [x] **Terminate the suspended VM → next message rehydrates with the Workspace and Agent session restored before ready, and model-side memory continuity across the replacement** — `checkpoint.test.ts` (round trip through a stand-in door over a real socket, Workspace outside HOME, CLI scratch excluded) and the **real-CLI probe below**: a second "VM" with a fresh HOME + Workspace restored the first's Checkpoint, resumed the pinned session, and its model call replayed the first VM's prompt and reply before the new Turn.
- [x] **A hard kill mid-Turn → the replacement's boot sweep terminalizes that Turn `interrupted`** — unchanged from #666/#669 (the awaited sweep in `configure()`); restore now precedes it, so history shows exactly what durably completed and the Workspace is the last suspend's.
- [x] **The Checkpoint pointer tracks the latest durable Checkpoint** — `postgres-conversation-vm-store.test.ts` (moves only for the row's VM, returns the previous key, survives terminate + re-claim) and `checkpoint.route.test.ts` (fresh key per PUT, previous object deleted, stale VM refused with its object removed).

## Verification

- `apps/in-vm-server`: `tsc --noEmit` clean, `bun run test` 117/117 (after the Ponytail trims)
- `apps/chat-api`: `bun run test` 227 pass / 4 skip (the real-Postgres lane); the one `tsc` error in `s3-artifact-download-signer.ts` is the pre-existing smithy drift on main
- `scripts/agent-maintenance-infrastructure.test.ts` 10/10; `terraform fmt -check` clean; `bunx biome check` clean
- GNU tar on `amazonlinux:2023-minimal` (after the image's `dnf install tar gzip`): the exact create/extract argument forms `checkpoint.ts` uses produce the two-root archive and restore both roots, excludes honoured.

**Real-CLI probe** (the pinned CLI 2.1.251 through the real `startDrainLoop` → `createAgentSession` → SDK `query()`, PGlite + the in-memory relay, a local fake Anthropic-Messages server recording every request; two sequential "VMs" with the image's fixed HOME/Workspace paths on fresh disk, a stand-in `/v2/checkpoint` door):

```
[probe] pinned session id 0a1b2c3d-4e5f-4a6b-8c7d-9e0f1a2b3c4d                # = the Conversation id
[probe] vm-a restore: none
[probe] vm-a turn-1 status: done
[probe] vm-a SDK message types: system:init:0a1b2c3d-…,…,result           # created under the pinned id
[probe] door PUT 6084 bytes { vm: "vm-a" }   checkpoint saved { elapsedMs: 39 }
[probe] vm-a transcript present for pinned id: true                     # ~/.claude/projects/<cwd-key>/<id>.jsonl
[probe] vm-b restore: restored
[probe] vm-b workspace after restore: [ "vm-a-note.md" ]                 # the Workspace file written on vm-a
[probe] vm-b turn-2 status: done
[probe] vm-b SDK message types: system:init:0a1b2c3d-…,…,result           # resumed under the same id
[probe] call 1: 2 messages; last user: "Remember this codeword: pelican. Reply briefly."   # vm-a's Turn
[probe] call 2: 5 messages; last user: What was the codeword?                              # vm-b's Turn, with vm-a's history
[probe] last call carries turn-1's prompt: true
[probe] last call carries vm-a's reply: true
[probe] CONTINUITY: PASS — vm-b's model call replays vm-a's Turn (prompt + reply) before turn-2
```

**Not run here — needs the real topology** (a scratch MicroVM image from this branch + this chat-api on Fargate; the auto-mode classifier blocks the ECR push and `run-task`): the platform actually firing `/suspend` and `/resume`, the 60 s hook budget against the real per-VM bandwidth, and the S3 `PutObject` streaming under the chat-api task role. The hand-launch section of `infra/microvm-image/README.md` gained the suspend → pointer → S3 → terminate → rehydrate steps for that pass. Until it runs, the platform's behaviour on a non-200 suspend hook (abort the suspend, or suspend anyway) is documented as unverified in the code.

## Review trail

Two Ponytail rounds so far: round 1 (eight findings: unused `fetch`/S3-client injection seams, a shared `admitGatewayCaller` for the gateway and checkpoint routes, a single `pause()` slot, folded tests, trimmed prose) applied in `887d7ad`; round 2 (five: the Conversation id is the session id with a UUID assert instead of a UUID-v5 helper, one suspend guard, no restore-time mkdirs, `.env.example` entry, docs paragraph) applied in `42db834`; round 3 (three: Bun's `$` shell instead of a `Bun.spawn` wrapper for `tar`, one `:hook{suspend|resume}` route, pointer-only file headers) applied in `d4c96ae`; round 4 (one restated doc sentence) in `b4e0ff1`; round 5 (four: one restore log line, no-op test loggers, a shorter hook comment and doc paragraph) in `12d30c8`; round 6 (three: `os.homedir()`, a void restore, the tautological `checkpoint-tools` smoke check) in `d9cd0a8`. A `/code-review` pass on `d9cd0a8` posted APPROVED FOR MERGE with non-blocking notes only; the substantive ones are folded in by the head commit: the pause promise now settles only when the loop parks idle (the doc's "lands after resume, when the drain completes" is true again), a suspend that finds a Turn queued during its hold refuses (500) instead of stranding it, the cap is 32 MiB so the low-end bandwidth can always restore inside the budget, `pause()` is reentrant, and the CONTEXT.md glossary is current (`34f9a15`). Ponytail rounds 8 and 9 each substituted `Promise.withResolvers` for a hand-rolled deferred (`feb53ef`, `a19799c`); round 10 on `a19799c`: NO FINDINGS. A second `/code-review` on `a19799c` posted APPROVED FOR MERGE; its non-blocking notes 0–3 (fail-closed VM id, the queued-behind-the-hold strand, the timed-out-hook hole, the flag reset) are folded in by `9914963`. Ponytail round 11 found two micro-trims (`6f56848`); round 12 on `6f56848`: NO FINDINGS. A third `/code-review` on `6f56848` posted APPROVED FOR MERGE (no blocking findings). CI on `6f56848` is green; its `test` job needed two re-runs past the documented PGlite WASM traps (a `RuntimeError: Unreachable code` in the untouched migration test, then `ERRORDATA_STACK_SIZE exceeded` in the untouched agent-worker package — both suites pass locally on this tree, 272/272 and 30/30).

**Follow-up material from the third review, deliberately not folded in here** (each would restart the review loop; none is blocking): single-flight the suspend handler so a hold outlived by a timed-out hook cannot run two Checkpoints concurrently; make the ship-together contract loud in the other direction too (a `/run` with a VM id but no `CHECKPOINT_URL`); the pre-existing #669 urgent-upgrade path terminates a *running* VM without a Checkpoint — gate it on `SUSPENDED`; a Workspace past the 32 MiB cap or a transcript the CLI refuses to `resume` wedges a Conversation with only a log line; the 60 s hook cap vs. a long Turn; `statsig/` under `~/.claude` may need excluding once the live pass shows what the idle CLI writes.

## Docs

`docs/agents/chat-api.md` gains "Checkpoint a v2 Conversation"; `configuration.md`, `architecture.md`, `security.md`, `database.md`, the image README, the topology runbook, and ADR-0034 (amendment) updated.

https://claude.ai/code/session_01CKW7eM1bBqbiUzwc2YgHV4


chengchao:
## Review of `d9cd0a8` (PR #700, closes #670 under spec #654)

Two-axis review (repository standards in `AGENTS.md` + `docs/agents/*.md`; conformance to #670 incl. its comments, #654, ADR-0034 + amendments, and the new "Checkpoint a v2 Conversation" section). Every finding below was verified against the code. Checks run: `apps/in-vm-server` `bun run test` 117/117 + `tsc --noEmit` clean; `apps/chat-api` `bun run test` 227 pass / 4 skip.

Verified OK (no finding): restore precedes the boot sweep and the drain loop (`index.ts` configure order); `swapCheckpointPointer` is one transaction, `FOR UPDATE`, guarded on `microvm_id`, and `claimLaunch` nulls `microvm_id` so a retired VM cannot pass during the claim → `RunMicrovm` → `recordLaunched` window; a failed PUT or a 409 never moves the pointer and the 409 path removes its object; PUT/GET stream (Node `Readable` into `PutObject`, `transformToWebStream` out); the chat-api task role has `s3:GetObject/PutObject/DeleteObject` on `conversations/*` (the route's delete would otherwise fail every PUT); the execution-role policy is gone per the ADR amendment; `suspendTimeoutInSeconds=60` is registered; the `userId` claim is required by the `(user_id, conversation_id)` PK and the ship-together break is called out in the PR body; nothing under `~/.claude` can revert VM policy (the deny bundle is `Options`-side and `/etc/claude-code/managed-settings.json` is outside the Checkpoint); `settingSources: []` ignores any restored `settings.json`.

### Blocking

None.

### Non-blocking

1. **Docs contradict the code on the post-timeout path.** `docs/agents/chat-api.md:366` says a Turn that outruns the suspend budget has "its Checkpoint land after resume, when the drain completes". Code: `apps/in-vm-server/src/drain-loop.ts:171-175` `resume()` calls `releaseParked()` unconditionally, so the still-pending `pause()` in `index.ts:179` settles at the `/resume` hook and `serving.checkpoint()` tars `~/.claude` + the Workspace **while the Turn is in flight**. Scenario: 20-min Turn, suspend fires at minute 5, hook times out, platform snapshots; on resume the tar runs against files the CLI is writing — GNU tar "file changed as we read it" exits 1, Bun `$` throws, the handler answers 500 to a request the platform abandoned minutes ago; or the tar succeeds and the pointer moves to a mid-Turn Checkpoint. Not data loss (the snapshot holds state, the previous pointer stays), but either fix the sentence or don't release the pause on resume.

2. **A Turn queued during the suspend hold is stranded until the next message.** `index.ts:143-148` `nudge` answers 200 while the loop is paused, the loop ignores the doorbell (`drain-loop.test.ts` "a nudge while paused claims nothing"), the hook answers 200, the platform snapshots with the row still `queued`. Nothing re-nudges: chat-api has no background work, and `ensure-vm.ts:143-145`'s "its interval self-heal drains the queue without this nudge" is now false during a pause (and frozen after it). Scenario: VM idle 15 min → suspend hook running (a few seconds) → user POSTs → SSE waits on `queued` until they send another message. Pre-PR the same race froze a *processing* Turn mid-stream, so this is not a regression — but a cheap guard fits here: after the Checkpoint lands, if `queued` rows exist for this Conversation, `resume()` and answer 500 (harmless if the platform suspends anyway; serves the Turn if it aborts). Worth a follow-up ticket with the ensure-vm comment fixed either way.

3. **Restore budget vs. size cap, and the wedge it can produce.** `checkpoint.ts:39` `TRANSFER_TIMEOUT_MS = 45_000` bounds the GET; `checkpoint.route.ts:15` caps a Checkpoint at 256 MiB, commented "the 60 s budget … cannot move more than this anyway" — 256 MiB in 45 s needs ~5.7 MB/s, while #670's comment quotes 1–16 MB/s per VM. Scenario at the low end: a Conversation whose Checkpoint passes ~45 MB restores past the timeout → `/run` 500 → platform terminates → nudge fails → `markTerminated` → the next message re-claims and hits the identical restore: a **permanent per-Conversation wedge** with no code path out (only a hand edit of `checkpoint_pointer`). Needs the live bandwidth number; until then either lower the cap to what 45 s can move, or give the restore a bounded escape (log + serve from empty is wrong for continuity, so a size check at *save* time is the better place).

4. **Suspend hold is unbounded against a 60 s hook.** `index.ts:179` awaits `loop.pause()` with no deadline — as AC2 literally asks — but #670's comment says the write "must complete within 60 s" and the PR admits the platform's behaviour on a timed-out suspend hook is unverified. If it terminates (as the run hook did on #666), a suspend mid-Turn loses everything since the previous Checkpoint and the Turn ends `interrupted` on rehydrate. Not fixable without knowing the platform's answer; this is the first thing the real-topology pass should establish (suspend a VM mid-Turn and see whether it snapshots or dies).

5. **`CONTEXT.md:154-161` glossary is stale** (`docs/agents/domain.md`: use the glossary's vocabulary): "written to the Conversation's S3 prefix by the In-VM server on the suspend **and terminate** hooks" — now brokered through chat-api, and terminate writes nothing (`app.ts:88`, `chat-api.md:366-367`). Update the entry in this PR; it is the one place the term is defined.

6. **`pause()` is not reentrant.** `drain-loop.ts:162-165` overwrites `parked`; a second `pause()` before the first settles leaves the first promise pending forever (its suspend handler never answers). Only reachable if the platform re-sends `/suspend` before `/resume`; a one-line "return the pending promise if one exists" closes it.

7. Nits: `app.ts:82` "so the platform never records a Checkpoint" — the platform records a *snapshot* (`CONTEXT.md` reserves the word); the `x-mymemo-microvm-id` fence is self-asserted by any valid-token holder (a retired VM's token lives up to 9 h) and holds only because it cannot learn its successor's id — worth one sentence in `security.md:28`.

APPROVED FOR MERGE

chengchao:
## Review of `a19799c` (PR #700, closes #670 under spec #654)

Two-axis review (repository standards in `AGENTS.md` + `docs/agents/*.md` + the `CONTEXT.md` glossary; conformance to #670 incl. its comment, #654, ADR-0034 + both amendments, and "Checkpoint a v2 Conversation"), with the `d9cd0a8..HEAD` follow-ups read line by line. Every finding was verified against the code. Checks run here: `apps/in-vm-server` `bun run test` 118/118 + `tsc --noEmit` clean; `apps/chat-api` `bun run test` 227 pass / 4 skip; plus a Bun 1.3 probe confirming a `Bun.file` PUT body carries `Content-Length` (not chunked), so the route's 411 gate never fires on the real client.

Verified OK on the new commits: `pause()` now settles only where the loop parks with nothing in flight (`drain-loop.ts:123,143`), and the first gate test proves it does not settle while a Turn streams; `pause()` is reentrant (one `PromiseWithResolvers`, `drain-loop.ts:172-179`); the post-Checkpoint refusal lifts the pause and answers 500 (`index.ts:186-197`); the 32 MiB cap is derived from the 45 s restore budget at 1 MB/s; the CONTEXT.md glossary, chat-api.md and security.md match the code; every prior non-blocking item except the two below was folded in as described. Unchanged and still correct: restore precedes the sweep and the loop; `swapCheckpointPointer` is `FOR UPDATE` + `microvm_id`-guarded and a refused PUT removes its object; task-role S3 grants include `DeleteObject`; the execution role carries no policy; `suspendTimeoutInSeconds=60` is registered.

### Blocking

None.

### Non-blocking

0. **The whole feature fails open when the `/run` body carries no `microvmId`.** `index.ts:69-72` builds the door only when `config.checkpointUrl && microvmId`; a run hook whose body lacks the field (or names it differently from the OpenAPI shape `app.ts:35` assumes — never observed live, since in-VM logs do not reach CloudWatch) yields `door = null`: no restore, no log, and `/suspend` answers 200 at once (`index.ts:176`), so the platform snapshots and a later terminate loses everything silently. The field matches `RunMicrovmResponse.microvmId` (`microvm-control-plane.ts:81`), so it is probably right — but it should be fail-closed: throw in `configure()` when `CHECKPOINT_URL` is set and `microvmId` is not (one assert), and the live pass should confirm the id is byte-equal to the one `recordLaunched` stored (a mismatch makes every PUT a 409 with only the `checkpoint from a retired VM refused` warn as evidence).

1. **The strand guard sees only nudges that arrive after `pause()`; a Turn queued before the suspend, behind the in-flight one, is still stranded.** `drain-loop.ts:71-72` sets `nudgedWhilePaused` only while `paused`. Scenario: Turn A running; the user sends B (chat-api enqueues + nudges, `paused=false`, so only `doorbell` is set); the platform suspends mid-A; A ends → `continue` → `doorbell=false` → paused branch parks and settles the pause; `nudgedWhilePaused()` is `false`; the Checkpoint lands, the hook answers 200, the platform snapshots with B `queued` until the next message. `drain-loop.test.ts:416-437` is exactly this state (`{ t1: "done", t2: "queued" }` at settle). The robust guard is a DB question, not a flag: one SELECT for `queued` rows on this Conversation after the Checkpoint (a `hasQueuedTurnsTx` beside `cancelQueuedTurnTx`), which also covers a nudge chat-api failed to deliver — `ensure-vm.ts:144-145` still says "its interval self-heal drains the queue without this nudge", which is false once the VM is snapshotted. Follow-up ticket material; the flag can stay as the fast path.

2. **"The Checkpoint is never taken mid-Turn" has one remaining hole on the timed-out-hook path.** A pause outlived by `/resume` settles at the idle park (`drain-loop.ts:143`) with `paused=false`, so `serving.checkpoint()` (`index.ts:185`) runs with the loop open: a nudge during the tar claims a Turn, and the tar runs against a transcript the CLI is writing (GNU tar exit 1 → 500 to a request the platform abandoned minutes ago; or a mid-Turn Checkpoint moves the pointer). Reachable only if the platform resumes a VM whose suspend hook timed out — the unverified case the PR names. Cheapest fix: in the suspend handler re-`pause()` after the first settle when `!paused` (or resolve `parked` only while `paused`, so a timed-out hook never checkpoints — arguably right, since the snapshot holds the state). `chat-api.md:365-367` overstates this sentence slightly.

3. **Reentrant `pause()` forgets a nudge between two suspends.** `drain-loop.ts:178` resets `nudgedWhilePaused = false` on every call, including one that returns the pending promise. A platform that re-sends `/suspend` before `/resume` loses the flag for a Turn queued between the two. One line: reset only inside the `if (!parked)` branch.

4. **No escape from a transcript the CLI refuses to `resume`.** `agent-session.ts:63-65`: transcript present → `resume` on every `start()`. If the CLI rejects it (a CLI pin bump that changes the transcript format; a truncated last line from item 2), the window ends result-less, the Turn ends `error`, `live` is retired, and the next Turn resumes the same file: a per-Conversation wedge with no code path out except deleting the Checkpoint by hand. Follow-up: a session that dies before `system:init` under `resume` should retire the transcript (rename) and start fresh under `sessionId`, logging the lost continuity.

5. **First things the real-topology pass must establish** (not verifiable here): (a) the run hook body's `microvmId` (`app.ts:35`, sent as `x-mymemo-microvm-id`) is byte-equal to `RunMicrovm`'s `microvmId` that `recordLaunched` stores (`microvm-control-plane.ts:81`) — any format difference makes every PUT a 409 and the Conversation silently un-checkpointable (the only trace is the `checkpoint from a retired VM refused` warn); (b) `PutObject` with a `Readable.fromWeb` body under Bun's `node:http` shim (the SDK's default flexible-checksum trailers) — this is chat-api's first streaming `PutObject`; (c) the 32 MiB cap is a permanent ceiling: a Workspace that outgrows it makes every later suspend `413` → 500 for the rest of the Conversation's life with only a log line — a size check at pack time (drop the Workspace's largest files, or fail with the size in the log) is the follow-up when the live bandwidth is measured.

6. Standards judgement calls (none hard): `checkpoint.route.ts:26,41` types the ref inline where `ConversationRef` exists; `checkpoint.test.ts:19-21` shares `objects`/`received`/`failWith` across `it` blocks, so the first test's `toHaveLength(1)` depends on order; `checkpoint.route.ts:65` `as never` where a `ReadableStream` cast would say what is bridged; `chat-api.md:350` "now `{ conversationId, userId, exp }`" dates the doc.

APPROVED FOR MERGE

chengchao:
## Review of `6f56848` (PR #700, closes #670 under spec #654)

Two-axis review (repository standards in `AGENTS.md` + `docs/agents/*.md` + the `CONTEXT.md` glossary; conformance to #670 incl. its comment, #654, ADR-0034 + both amendments, "Checkpoint a v2 Conversation"), with `a19799c..HEAD` read line by line for anything new. Every finding was verified against the code. Checks run here: `apps/in-vm-server` `bun run test` 117/117 + `tsc --noEmit` clean; `apps/chat-api` `bun run test` 227 pass / 4 skip; `packages/agent-db/src/turn-store.test.ts` 23/23.

Verified OK on the new commits: `configure()` fails closed when `CHECKPOINT_URL` names a door but `/run` names no VM (`index.ts:76-80`); `hasQueuedTurnsTx` is Conversation-scoped through `conversationTurns` (`turn-store.ts:45-51,220-231`) and runs *after* the Checkpoint lands, so a refusal never loses state; `releaseParked()` is now reached only inside `if (paused)` (`drain-loop.ts:117-122`), so a hold outlived by `/resume` cannot settle mid-Turn or at an open idle park, and the last gate test pins exactly that; concurrent `PUT`s cannot leave the pointer dangling — each key is deleted at most once and only after it was replaced under the `FOR UPDATE` swap (`postgres-conversation-vm-store.ts:106-124`); `claimLaunch` nulls `microvm_id` but keeps `checkpoint_pointer` (`:25-31`), so a retired VM's late `PUT` is a 409 and a rehydrate finds the pointer; the `conversationId` in the object key is token-verified, never the raw path; no standards breach (Drizzle peer sets, store ownership, secret rule, glossary all hold).

### Blocking

None.

### Non-blocking

1. **New in 9914963 — an outlived hold makes the next suspend run two Checkpoints concurrently.** `drain-loop.ts:166` `parked ??=` hands the *same* promise to a stale `/suspend` handler (one whose hook timed out mid-Turn, then `/resume` came) and to the next `/suspend`; both settle at the next paused park and both run `serving.checkpoint()` (`index.ts:198`): two tars + two `PUT`s sharing the 45 s budget (`checkpoint.ts:38`). Sharper variant: a nudge lands during the tars; whichever handler checks `hasQueuedTurns()` first answers `resume()` + 500 (`index.ts:199-209`), the loop claims the Turn, and the *other* handler's tar now reads a transcript the CLI is writing — GNU tar exit 1 → 500 (fine), or a torn transcript lands as the pointed Checkpoint and its `hasQueuedTurns` sees `processing`, not `queued`, so it answers 200 and the platform snapshots. Reachable only on the unverified timed-out-hook-then-resume path. Cheap fix: single-flight the suspend handler in `index.ts` (`let suspending: Promise<void> | null`; a second `/suspend` awaits the one in flight).
2. **Urgent upgrade terminates a *running* VM with no Checkpoint** (`ensure-vm.ts:175-191`, pre-existing #669, flag default off): #654 says the flag "converts the next **resume** into a rehydrate", i.e. a suspended VM whose suspend-time Checkpoint is complete. As written, a VM that is live (idle < 15 min, or mid-Turn) is killed and the delta since its last Checkpoint — possibly the whole Conversation if it never suspended — is gone, breaking AC3's memory continuity on that path. Gate on `GetMicrovm` state `SUSPENDED` (or nudge normally and let the next idle suspend land the upgrade). Follow-up ticket.
3. **The strand check has a residual window**: a Turn enqueued + nudged after `hasQueuedTurns()` returns false (`index.ts:199`) but before the platform snapshots is answered 202 by a paused loop (`ensure-vm.ts:129`, `drain-loop.ts:117`) and waits for the next POST. Milliseconds wide, not closable from the VM; noting it so the live pass does not mistake it for a bug elsewhere.
4. **Fail-closed is one-directional**: `index.ts:76` throws for door-without-VM but a `/run` with `microvmId` and no `CHECKPOINT_URL` (older chat-api payload, or a hand-typed runbook payload that omits it) serves doorless: no restore, no save, only silence. One more assert (`microvmId && !config.checkpointUrl`) would make the ship-together contract loud.
5. **`tar` exit 1 fails the suspend.** `checkpoint.ts:53` is `$…quiet()` with no `.nothrow()`: GNU tar exits 1 (archive still valid) on "file changed as we read it". The idle CLI process stays alive between Turns; if it touches anything under `~/.claude` outside `debug`/`shell-snapshots` during the pack (e.g. `statsig/`), the suspend answers 500 and the VM stays up until the platform retries. Worth a `.nothrow()` + `exitCode > 1` check, or adding `statsig` to `CLAUDE_EXCLUDES`, once the live pass shows what the idle CLI writes.
6. **Still open from the earlier passes (follow-up ticket material, not re-argued):** a Turn longer than the 60 s hook cap leaves the Checkpoint unwritten and the platform's behaviour unverified (`index.ts:192`); the Workspace is unbounded so a Conversation past 32 MiB gets `413` → 500 on every suspend for life (`checkpoint.route.ts:20,60`); the run hook must fit GET (≤ 45 s) + unpack + exec-verify + sweep in 60 s through a cold ENI (`index.ts:88-140`); a Conversation active every < 15 min for 8 h never suspends and the at-cap kill loses it entirely (#654 defers pre-cap rotation); a transcript the CLI refuses to `resume` wedges the Conversation (`agent-session.ts:63-65`).
7. Nits: `database.md:23` / `schema.ts:267` say "one write guarded on the VM id" — the guard is on the `FOR UPDATE` select, the `UPDATE` itself is by PK (`postgres-conversation-vm-store.ts:118-121`; correct under the lock, and adding `microvm_id` to the `UPDATE` predicate is a one-liner that makes the doc literally true); `CONTEXT.md` **Gateway** still says VM egress is locked to "RDS, Redis, and this route" — the door is now the ALB (two routes); `index.ts:151` passes the whole `TurnServingDeps` bag as the `{ userId, conversationId }` key where `:137-140` spells it out; `turn-store.test.ts:153-161` never enqueues under a second Conversation, so the scoping of `hasQueuedTurnsTx` is unpinned.

APPROVED FOR MERGE

